PR #25, opened 2025-10-27
fix: make integration tests fail fast and fix validation error handling

yilongli:
Problem: Integration tests currently loop forever when validation fails instead of terminating with a clear error message.

Changes:
- Add stop_on_first_err flag to ChainSyncConfig to terminate chain_sync on first error (enabled in integration tests, disabled in production)
- Fix grow_local_chain to use error_message from ValidationResult instead of constructing incorrect StateRootMismatch errors where actual and claimed fields were the same
- Change FailedValidation variant to accept String for better error messages
- Fix clippy warnings (ok_or_else -> ok_or)


flyq (approved):
lgtm

PR #26, opened 2025-10-28
opt: Optimize remote chain tracking and batch validation task ingestion.

flyq:
- Increase the remote tracker lookahead window to 80 blocks and tighten the poll interval to 100 ms so the validator keeps a healthier buffer of remote data.
  - Rework the tracker loop to fetch headers plus Salt/MPT witnesses in parallel, stop at the first RPC/DB failure, and queue the successful batch once the join completes.
  - Replace the single-task insert with ValidatorDB::add_validation_tasks, batching block data, both witness types, and task queue entries in one transaction to cut write overhead.
  - Extend grow_remote_chain to accept an iterator of headers, appending a validated sequence in a single transaction while preserving parent-link checks.

### DIFF
--- a/bin/stateless-validator/src/main.rs
+++ b/bin/stateless-validator/src/main.rs
@@ -457,7 +457,8 @@ async fn remote_chain_tracker(
                 remote_tip.0 + 1
             );
 
-            let fetch_data = future::join_all(
+            // Fetch blocks in parallel
+            let tasks = future::join_all(
                 (remote_tip.0 + 1..remote_tip.0 + 1 + blocks_to_fetch).map(|block_number| {
                     let client = client.clone();
                     tokio::spawn(async move {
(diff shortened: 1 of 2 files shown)


bin/stateless-validator/src/main.rs
```diff
@@ -110,8 +112,8 @@ impl Default for ChainSyncConfig {
             concurrent_workers: num_cpus::get(),
             sync_poll_interval: Duration::from_secs(1),
             sync_target: None,
-            tracker_lookahead_blocks: 10,
-            tracker_poll_interval: Duration::from_secs(2),
+            tracker_lookahead_blocks: 80,
```
yilongli, line 115: 80这个值是怎么定的？

flyq: 之前设置40的时候，还有部分情形会一次请求40个block，说明gap为0，workers 在等待。就设置了一个80值

yilongli (approved):

PR #27, opened 2025-10-31
chore(deps): update salt deps and adapt to new Witness::create() API

yilongli:
Update salt dependency to rev 57fa05e. Refactor state update logic in executor.rs to adapt to the new Witness::create() API by separating updates from inserts/deletes and applying updates first, then inserts/deletes in deterministic order to match the witness's expected key ordering.


flyq (approved):
lgtm

PR #29, opened 2025-11-01
fix: add bytecode codehash validation and improve integration test

yilongli:
Add validation to ensure RPC provider returns bytecode matching the expected codehash from the witness. This prevents malicious or corrupted bytecode from being used during validation.

The integration test is updated to properly exercise bytecode fetching by removing contract pre-population and implementing an eth_getCode RPC handler in the mock server. This makes the test more effective by testing the actual code path that fetches and validates bytecode.

### DIFF
--- a/bin/stateless-validator/src/main.rs
+++ b/bin/stateless-validator/src/main.rs
@@ -1,8 +1,8 @@
 use alloy_genesis::Genesis;
 use alloy_primitives::{Address, B256, BlockHash, BlockNumber, hex};
-use alloy_rpc_types_eth::{Block, BlockId};
+use alloy_rpc_types_eth::{Block, BlockId, BlockNumberOrTag};
 use clap::Parser;
-use eyre::{Result, anyhow};
+use eyre::{Result, anyhow, ensure};
 use futures::future;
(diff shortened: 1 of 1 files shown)


flyq (approved):
lgtm

bin/stateless-validator/src/main.rs
```diff
@@ -603,11 +601,19 @@ async fn validate_one(
 
             // Fetch missing contract codes via RPC and update the local DB
             let codes = client
-                .get_code(&missing_contracts, (block_number - 1).into())
+                .get_code(&missing_contracts, BlockNumberOrTag::Latest)
```
gjaldon, line 604: This can cause an issue when [`block_number`](https://github.com/megaeth-labs/stateless-validator/blob/main/bin/stateless-validator/src/main.rs#L678) isn't the latest block number. If there are changes to a contracts' bytecode between `block_number` and latest block, the incorrect bytecode will be fetched. This seems only possible via self-destruct though.

Any bytecode mismatches will be [caught](https://github.com/megaeth-labs/stateless-validator/blob/main/bin/stateless-validator/src/main.rs#L701-L705) but can stall the stateless validator.

PR #58, opened 2025-12-23
Fix Graceful Shutdown to Prevent Validator Freeze

flyq:
### Problem

There was a low probability that running `systemctl stop/restart stateless-validator` would cause the system to freeze, eventually get killed after 60 seconds (systemd's default timeout).

**Root Cause:**
1. When using `#[tokio::main]`, the runtime would wait indefinitely for `spawn_blocking` tasks to complete during shutdown, causing the process to hang until systemd forcefully killed it with `SIGKILL`.
2. The service only handled `SIGINT` (Ctrl+C), but `systemctl stop/restart` sends `SIGTERM`. Since `SIGTERM` wasn't being caught, the application didn't initiate a graceful shutdown.

### Solution

1. **Added SIGTERM handler**: The application now properly handles both `SIGINT` and `SIGTERM` signals, ensuring `systemctl stop/restart` triggers a graceful shutdown.

2. **Manual runtime management with shutdown timeout**: Replaced `#[tokio::main]` with explicit `Runtime::new()` and `rt.shutdown_timeout(Duration::from_secs(5))`. This ensures the runtime will force-stop any lingering `spawn_blocking` tasks after 5 seconds.

### Changes

- `bin/stateless-validator/src/main.rs`: 
  - Split `main()` into sync `main()` and `async_main()`
  - Added manual tokio runtime with 1-second shutdown timeout
  - Added `SIGTERM` signal handler alongside existing `SIGINT` handler

### Testing

- Verified `systemctl stop stateless-validator` now exits cleanly within seconds
- Verified `systemctl restart stateless-validator` completes without hanging
- Confirmed database integrity is preserved after stop/restart cycles

### DIFF
--- a/bin/stateless-validator/src/main.rs
+++ b/bin/stateless-validator/src/main.rs
@@ -235,7 +235,6 @@ fn main() -> Result<()> {
 
     let rt = Runtime::new()?;
     let result = rt.block_on(async_main());
-
     // Force shutdown after 5 seconds to avoid hanging on spawn_blocking tasks
     rt.shutdown_timeout(Duration::from_secs(5));
 
@@ -333,7 +332,7 @@ async fn async_main() -> Result<()> {
     let validator_logic = chain_sync(client.clone(), validator_db.clone(), config, chain_spec);
 
     let mut sigterm = signal::unix::signal(signal::unix::SignalKind::terminate())
-        .expect("Failed to register SIGTERM handler");
+        .map_err(|e| anyhow!("[Main] Failed to register SIGTERM handler: {e}"))?;
     tokio::select! {
         res = validator_logic => res?,
         _ = signal::ctrl_c() => {
--- a/bin/stateless-validator/src/main.rs
+++ b/bin/stateless-validator/src/main.rs
@@ -14,7 +14,7 @@ use futures::future;
 use op_alloy_rpc_types::Transaction;
 use revm::{primitives::KECCAK_EMPTY, state::Bytecode};
 use salt::SaltWitness;
-use tokio::{runtime::Runtime, signal, task};
+use tokio::{signal, task};
 use tracing::{debug, error, info, warn};
 use tracing_subscriber::{EnvFilter, Layer, fmt, layer::SubscriberExt, util::SubscriberInitExt};
 use validator_core::{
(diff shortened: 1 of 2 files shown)


bin/stateless-validator/src/main.rs
```diff
@@ -322,10 +332,15 @@ async fn main() -> Result<()> {
 
     let validator_logic = chain_sync(client.clone(), validator_db.clone(), config, chain_spec);
 
+    let mut sigterm = signal::unix::signal(signal::unix::SignalKind::terminate())
+        .expect("Failed to register SIGTERM handler");
```
yilongli: `.map_err(|e| anyhow!("Failed to register SIGTERM handler: {e}"))?;`

bin/stateless-validator/src/main.rs
```diff
@@ -230,9 +230,19 @@ struct CommandLineArgs {
     metrics_port: u16,
 }
 
-#[tokio::main]
-async fn main() -> Result<()> {
+fn main() -> Result<()> {
```
yilongli, line 233: 是不是只需要保留335-343行的改动就能解决`systemctl stop`命令导致freeze的问题了

flyq: 还需要rt.shutdown_timeout，因为 freeze 主要是 spawn_blocking 导致，然后通过 `rt.shutdown_timeout(Duration::from_secs(5));` 来关闭。

这个是 shutdown_timeout 的文档：
```rust
    /// Shuts down the runtime, waiting for at most `duration` for all spawned
    /// work to stop.
    ///
    /// See the [struct level documentation](Runtime#shutdown) for more details.
    ///
    /// # Examples
    ///
    /// ```
    /// # #[cfg(not(target_family = "wasm"))]
    /// # {
    /// use tokio::runtime::Runtime;
    /// use tokio::task;
    ///
    /// use std::thread;
    /// use std::time::Duration;
    ///
    /// fn main() {
    /// #  if cfg!(miri) { return } // Miri reports error when main thread terminated without waiting all remaining threads.
    ///    let runtime = Runtime::new().unwrap();
    ///
    ///    runtime.block_on(async move {
    ///        task::spawn_blocking(move || {
    ///            thread::sleep(Duration::from_secs(10_000));
    ///        });
    ///    });
    ///
    ///    runtime.shutdown_timeout(Duration::from_millis(100));
    /// }
    /// # }
    /// ```
    pub fn shutdown_timeout(mut self, duration: Duration) {
        // Wakeup and shutdown all the worker threads
        self.handle.inner.shutdown();
        self.blocking_pool.shutdown(Some(duration));
    }
```

yilongli (approved):

PR #59, opened 2025-12-24
Fix: Add Anchor Block for Validation Reporting

flyq:
### Summary

This PR introduces an immutable anchor block mechanism to provide a stable starting point for validation reporting. Previously, the validation reporter used `get_earliest_local_block()` which could change over time as blocks are pruned. This caused issues when the upstream validator's last validated block fell below the local chain's earliest block after pruning.

#### New `ANCHOR_BLOCK` Table
- Added a new database table `ANCHOR_BLOCK` that stores the initial trusted starting block `(BlockNumber, BlockHash)`
- The anchor is set once during chain reset and remains immutable
- Provides a stable reference point for validation gap detection

#### Updated Validation Reporter
- Changed `get_earliest_local_block()` → `get_anchor_block()` when determining chain bounds
- Fixed log ordering: error message now logs before returning on validation gap detection

### Files Changed

| File | Changes |
|------|---------|
| `validator-core/src/validator_db.rs` | Added `ANCHOR_BLOCK` table definition, `get_anchor_block()` method, and anchor persistence in `prune_history()` |
| `bin/stateless-validator/src/main.rs` | Updated reporter to use anchor block; fixed error log placement |

### Why This Matters

The anchor block ensures that validation gap detection works correctly even as the local chain is pruned. The reporter always knows the original starting point of validation, preventing false validation gap errors.

### DIFF
--- a/bin/stateless-validator/src/main.rs
+++ b/bin/stateless-validator/src/main.rs
@@ -884,16 +884,16 @@ async fn validation_reporter(
             Ok(response) => {
                 // Check for validation gap
                 if response.last_validated_block.0 < first_block.0 {
-                    error!(
-                        "[Reporter] Report rejected for blocks {first_block:?}-{last_block:?}, upstream at {:?}",
-                        response.last_validated_block
-                    );
                     return Err(anyhow!(
                         "Validation gap detected: upstream at block {}, but local chain starts at {}. Cannot advance validation.",
                         response.last_validated_block.0,
                         first_block.0
                     ));
                 }
+                error!(
+                    "[Reporter] Report rejected for blocks {first_block:?}-{last_block:?}, upstream at {:?}",
+                    response.last_validated_block
+                );
             }
             Err(e) => {
                 error!("[Reporter] Failed to report blocks: {e}");
--- a/bin/stateless-validator/src/main.rs
+++ b/bin/stateless-validator/src/main.rs
@@ -290,10 +290,6 @@ async fn run() -> Result<()> {
             .map_err(|e| anyhow!("Failed to fetch block {}: {}", block_hash, e))?;
 
         validator_db
-            .set_start_block(block.header.number, block.header.hash)
-            .map_err(|e| anyhow!("Failed to set start block: {}", e))?;
-
-        validator_db
             .reset_anchor_block(
                 block.header.number,
                 block.header.hash,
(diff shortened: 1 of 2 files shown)


bin/stateless-validator/src/main.rs
```diff
@@ -289,6 +289,10 @@ async fn run() -> Result<()> {
             .await
             .map_err(|e| anyhow!("Failed to fetch block {}: {}", block_hash, e))?;
 
+        validator_db
+            .set_start_block(block.header.number, block.header.hash)
```
yilongli: 这个函数和底下297行的`reset_anchor_block`重合了

validator-core/src/validator_db.rs
```diff
@@ -196,6 +196,16 @@ const CONTRACTS: TableDefinition<[u8; 32], Vec<u8>> = TableDefinition::new("cont
...
 
+/// Stores the immutable starting block for validation reporting.
+///
+/// **Schema:** Maps a singleton key (&str) to (BlockNumber, BlockHash) as (u64, [u8; 32])
+/// - Key: String "start_block" as &str
+/// - Value: (Block number as u64, Block hash as [u8; 32])
+///
+/// This table stores the initial trusted block that validation started from.
+/// Used by the validation reporter to report a stable first block.
+const START_BLOCK: TableDefinition<&str, (u64, [u8; 32])> = TableDefinition::new("start_block");
```
yilongli: `START_BLOCK` => `ANCHOR_BLOCK`
我们已经有了后者，没必要再引入一个新的概念。

validator-core/src/validator_db.rs
```diff
@@ -426,6 +437,51 @@ impl ValidatorDB {
...
+    /// initialization. Once set, this value should not be modified.
+    ///
+    /// # Arguments
+    /// * `block_number` - The block number of the starting block
+    /// * `block_hash` - The hash of the starting block
+    ///
+    /// # Returns
+    /// * `Ok(())` - Start block successfully stored
+    /// * `Err(ValidationDbError)` - Database error
+    pub fn set_start_block(
```
yilongli: 这个方法并入`reset_anchor_block`

validator-core/src/validator_db.rs
```diff
@@ -426,6 +437,51 @@ impl ValidatorDB {
...
+
+    /// Retrieves the immutable starting block for validation reporting
+    ///
+    /// Returns the trusted starting block.
+    ///
+    /// # Returns
+    /// * `Ok(Some((block_number, block_hash)))` - Start block found
+    /// * `Ok(None)` - No start block has been set
+    /// * `Err(ValidationDbError)` - Database error
+    pub fn get_start_block(&self) -> ValidationDbResult<Option<(BlockNumber, BlockHash)>> {
```
yilongli: 改名为`get_anchor_block`

yilongli (approved):

PR #71, opened 2026-02-01
fix: variable name

meelon-dev:
The [README](https://github.com/meelon-dev/stateless-validator/blob/main/README.md)  file refers the following variables: 
- `STATELESS_VALIDATOR_LOG_FILE_FILTER`
- `STATELESS_VALIDATOR_LOG_STDOUT_FILTER`

 However, further down in the same file, they are used as:
- `STATELESS_VALIDATOR_LOG_FILE` (without `_FILTER`)
- `STATELESS_VALIDATOR_LOG_STDOUT` (without `_FILTER`)

This [file also](https://github.com/meelon-dev/stateless-validator/blob/main/bin/stateless-validator/src/main.rs) uses:
- `STATELESS_VALIDATOR_LOG_FILE` (without `_FILTER`)
- `STATELESS_VALIDATOR_LOG_STDOUT` (without `_FILTER`)

Removed `_FILTER` from the variables, which does not break the code examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,8 @@ cargo build --release
 
 ```bash
 STATELESS_VALIDATOR_LOG_FILE_DIRECTORY=/path/to/log_dir \
-STATELESS_VALIDATOR_LOG_FILE_FILTER=debug \
-STATELESS_VALIDATOR_LOG_STDOUT_FILTER=info \
+STATELESS_VALIDATOR_LOG_FILE=debug \
+STATELESS_VALIDATOR_LOG_STDOUT=info \
 cargo run --bin stateless-validator -- \
   --data-dir /path/to/validator/data \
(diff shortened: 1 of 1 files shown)


Troublor:
Thanks for pointing out the discrepancy. We are refactoring the logging logic and will fix this along with the refactoring.

flyq:
Thanks for the fix! Closing this as the inconsistency no longer exists on `main`: the whole `STATELESS_VALIDATOR_LOG_*` env-var family was renamed to the shared `STATELESS_LOG_*` prefix when logging was extracted into `stateless-common` (#74), and the legacy-name migration layer was removed in #127. The current README and code now consistently use `STATELESS_LOG_FILE` / `STATELESS_LOG_STDOUT` (see `crates/stateless-common/src/logging.rs`), so this patch no longer applies.

PR #77, opened 2026-02-07
Add dedicated `report_validation_endpoint` and optimize block header fetching

flyq:
## Summary

- **Add dedicated `--report-validation-endpoint` CLI argument** replacing the boolean `--report-validation-results` flag, allowing a separate RPC endpoint to be configured specifically for reporting validated blocks via `mega_setValidatedBlocks`. A new `report_provider` is introduced in `RpcClient` so that validation reports are sent to a dedicated endpoint rather than reusing the data provider.
- **Optimize block header fetching** by introducing a `get_header()` method that uses `eth_getHeaderByNumber` / `eth_getHeaderByHash` instead of `eth_getBlockByNumber(…, false)`. This avoids transferring the full transaction hash list when only the header is needed, which is significant at high TPS. All call sites that previously fetched a full block just for header fields (number, hash, state_root, withdrawals_root) now use the lighter header-only RPC calls.
- **Remove unused dependencies**: `alloy-rlp`, `alloy-signer`, `revm-inspectors`, `alloy-provider`, `alloy-serde`, `op-alloy-network`.

## Changes

| File | Description |
|------|-------------|
| `crates/validator-core/src/rpc_client.rs` | Add `report_provider` field, `get_header()` method with hash verification, refactor `get_block_hash()` to delegate to `get_header()`, route `set_validated_blocks` through the dedicated report provider |
| `bin/stateless-validator/src/main.rs` | Replace `--report-validation-results` (bool) with `--report-validation-endpoint` (URL), pass report endpoint to `RpcClient`, update all `get_block()` header-only calls to `get_header()` |
| `bin/debug-trace-server/src/main.rs` | Pass `None` for the new `report_api` parameter, replace `get_block()` with `get_header()` for anchor block initialization |
| `bin/debug-trace-server/src/data_provider.rs` | Replace `get_block()` with `get_header()` / `get_block_hash()` for header-only lookups |
| `crates/validator-core/src/chain_sync.rs` | Replace `get_block()` with `get_header()` / `get_block_hash()` in sync logic |
| `README.md` | Update CLI docs and environment variable references for the new endpoint argument |
| `Cargo.lock`, `*/Cargo.toml` | Remove unused dependencies |

## Test plan

- [x] Verify the validator starts correctly with `--report-validation-endpoint <url>` and reports validated blocks to the specified endpoint
- [x] Verify the validator starts correctly without `--report-validation-endpoint` (reporting disabled)
- [x] Verify block header fetching works correctly via the new `get_header()` method
- [x] Verify the debug-trace-server still initializes and operates correctly with the updated header-only calls
- [x] Run existing unit tests (`cargo test`) to confirm no regressions


crates/validator-core/src/rpc_client.rs
```diff
@@ -283,18 +295,76 @@ impl RpcClient {
...
+        }
+
+        // Verify header hash matches the computed hash
+        ensure!(
+            header.hash_slow() == header.hash,
+            "Header hash mismatch: expected {:?}, computed {:?}",
+            header.hash,
+            header.hash_slow()
+        );
+
```
Troublor: It is unnecessary here to verify the header matches its hash, whose calculation is quite slow. Is there any specific reason or concern that a header may have an incorrect hash?

crates/validator-core/src/rpc_client.rs
```diff
@@ -367,14 +437,15 @@ impl RpcClient {
...
+    /// Reports a range of validated blocks via the dedicated report endpoint.
     pub async fn set_validated_blocks(
         &self,
         first_block: (u64, B256),
         last_block: (u64, B256),
     ) -> Result<SetValidatedBlocksResponse> {
-        let result = self
-            .data_provider
+        let provider =
+            self.report_provider.as_ref().ok_or_else(|| eyre!("Report provider not configured"))?;
```
Troublor, line 487: If we report error here, then we should not allow report_provider to be optional. Otherwise, we should fallback to the data_provider if report_provider is not present.

crates/validator-core/src/rpc_client.rs
flyq: validator 要求只信任起始时传入的 start block hash，并且检验所有从上游拉取的数据。至少在启动并传入 start block hash 时，需要校验这个head内容和 block hash 的是否对应，这样才能信任header 里面的 state root，withdrawal root 等并写入本地 canonical chain
我加一个是否校验的 bool 参数，然后 review 一下有哪些场景不需要校验

Troublor: 好的，加一个参数来选择性地校验吧，至少rpc server是不需要校验的。

crates/validator-core/src/rpc_client.rs
flyq: 嗯，这里感觉直接停机比较好。代码走到这里，说明一定提供了 report_provider 的 endpoint，因为是否 report 就是根据参数里面的 report_provider endpoint 是否是 some。只能说参数提供错了，停机重新启动比较好。
data_provider 目前就是直接连接普通rpc，肯定是没有 mega_setValidatedBlock 这个 methods 的

crates/validator-core/src/rpc_client.rs
flyq: fixed

Troublor (approved):

krabat-l (approved):

PR #103, opened 2026-03-23
perf: remove unnecessary clone in transaction execution loop

KENILSHAHH:
## Summary                                                                                              
                                                                                                          
  `execute_transactions` previously called `tx.inner.clone().into_inner()` on every                       
  transaction to obtain an owned `T`, solely to take a reference `&T` from it.                            
  This resulted in a heap allocation and full copy of the transaction data on every                       
  iteration of the hot execution loop, with the allocation being immediately discarded                    
  at the end of each loop body.
                                                                                                          
  The fix replaces this with `tx.inner.inner.as_recovered_ref()`, which converts
  `&Recovered<T>` to `Recovered<&T>` in-place by wrapping a pointer to the existing                       
  data.                                                     
  No allocation or copy is performed.                                                                     
   
  The `T: Clone` bound on `execute_transactions` is also removed, as it existed                           
  solely to support the now-removed clone.                  

  **Note:** The tracing executor (`tracing_executor.rs`) already accessed transaction                     
  data without cloning — this change brings `execute_transactions` in line with that
  existing pattern.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -77,7 +77,7 @@ jobs:
           additional_permissions: |
             actions: read
           claude_args: |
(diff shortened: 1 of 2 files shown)


flyq (approved):
good job!

KENILSHAHH:
adding some more changes through another branch @flyq

PR #106, opened 2026-03-24
perf: remove unnecessary clone in transaction execution loop

KENILSHAHH:
## Summary                                                                                              
                                                                                                          
  `execute_transactions` previously called `tx.inner.clone().into_inner()` on every                       
  transaction to obtain an owned `T`, solely to take a reference `&T` from it.                            
  This resulted in a heap allocation and full copy of the transaction data on every                       
  iteration of the hot execution loop, with the allocation being immediately discarded                    
  at the end of each loop body.
                                                                                                          
  The fix replaces this with `tx.inner.inner.as_recovered_ref()`, which converts
  `&Recovered<T>` to `Recovered<&T>` in-place by wrapping a pointer to the existing                       
  data.                                                     
  No allocation or copy is performed.                                                                     
   
  The `T: Clone` bound on `execute_transactions` is also removed, as it existed                           
  solely to support the now-removed clone.                  

  **Note:** The tracing executor (`tracing_executor.rs`) already accessed transaction                     
  data without cloning — this change brings `execute_transactions` in line with that
  existing pattern.

### DIFF
--- a/crates/stateless-core/src/executor.rs
+++ b/crates/stateless-core/src/executor.rs
@@ -601,4 +601,38 @@ mod tests {
             .unwrap_or_else(|e| panic!("validate_block failed for {number} ({hash}): {e:?}"));
         }
     }
+
+    /// Verifies that `replay_block` (which exercises `execute_transactions`) produces
+    /// receipts root, logs bloom, and gas used that match the block header.
+    #[test]
+    fn replay_block_output_matches_header() {
+        let fx = TestFixtures::mainnet();
+        let chain_spec = ChainSpec::from_genesis(fx.load_genesis().unwrap());
+        let paired = fx.paired_blocks();
+        let (_num, hash) = paired.first().expect("no paired mainnet fixtures");
+
+        let block = &fx.blocks[hash];
+        let salt_witness = fx.salt_witnesses[hash].clone();
+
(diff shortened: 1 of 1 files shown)


flyq (dismissed):
I rebaseed the main branch, and local e2e tests are all normal.
LGTM

flyq (approved):

flyq (approved):

Troublor (approved):

flyq:
Thanks for this PR — it correctly flagged the per-transaction `clone().into_inner()` in the hot execution loop and put the tx-clone cost on our radar. That awareness carried into the `BlockInput` refactor (#141), which landed the same approach this PR proposed: `txs_recovered()` now borrows each transaction in place via `tx.inner.inner.as_recovered_ref()` (see `crates/stateless-core/src/executor.rs`), and `execute_transactions` takes an `Iterator<Item = Recovered<&OpTxEnvelope>>` with no `Clone` bound. So while this patch no longer applies to the current code, the issue it identified is fixed — closing as superseded by #141.

KENILSHAHH:
thanks @flyq

batatmosfer-byte (commented):
Kluczowe, w której gałęzi jest klon gdzie szukać

PR #107, opened 2026-03-24
perf: stack-allocate PlainKey encoding in hot EVM state read path

KENILSHAHH:
### Stack-allocate PlainKey encoding in hot EVM state read path
                                                                                                          
  - Add `PlainKey::encode_account()` and `PlainKey::encode_storage()` returning
    fixed-size stack arrays (`[u8; 20]` and `[u8; 52]`) instead of heap-allocated `Vec<u8>`               
  - Update the four hot-path callers in `database.rs` (`basic_ref`, `storage_ref`,
    `bucket_id_for_account`, `bucket_id_for_slot`) to use the new stack-allocating methods                
  - Eliminate a heap allocation on every EVM state read during block execution
                                                                                                          
  Account and storage key sizes are fixed by the Ethereum specification (20 and 52 bytes
  respectively), making stack arrays the correct choice. The existing `encode() -> Vec<u8>`               
  is retained for the `executor.rs` use-site where an owned allocation is still required.

### DIFF
--- a/crates/stateless-core/src/data_types.rs
+++ b/crates/stateless-core/src/data_types.rs
@@ -71,6 +71,19 @@ impl PlainKey {
         }
     }
 
+    /// Encodes an account address into a stack-allocated 20-byte array.
+    pub fn encode_account(addr: Address) -> [u8; ACCOUNT_ADDRESS_LEN] {
+        addr.into_array()
+    }
+
+    /// Encodes an address + storage slot into a stack-allocated 52-byte array.
+    pub fn encode_storage(addr: Address, slot: B256) -> [u8; STORAGE_SLOT_KEY_LEN] {
+        let mut buf = [0u8; STORAGE_SLOT_KEY_LEN];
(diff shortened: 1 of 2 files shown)


KENILSHAHH:
Hello @flyq @Troublor I'd appreciate your feedback on this.

KENILSHAHH:
Rebased with main

flyq:
Thanks for this PR — it correctly identified the per-state-read heap allocation in the hot EVM read path (`PlainKey::encode()` returning a `Vec<u8>` on every `basic_ref`/`storage_ref`/bucket-id lookup). The same fix is landing via #170, which adds allocation-free encodings — `account_key_bytes(&Address) -> &[u8]` (a direct borrow) and `storage_key_bytes(Address, B256) -> FixedBytes<52>` (stack-allocated) — and switches all four hot-path call sites over, keeping `encode() -> Vec<u8>` only where an owned buffer is genuinely needed. The original patch no longer applies cleanly since `database.rs` was renamed to `evm_database.rs`, so closing this as superseded by #170.

PR #110, opened 2026-03-31
feat: add per-method RPC timeout configuration

KENILSHAHH:
## Summary

  - Previously all RPC calls shared no timeout, making it impossible to balance
    the tradeoff between witness fetches (which can legitimately take several
    seconds on heavy blocks) and block/header fetches (which should return in
    milliseconds). A single tight timeout caused false witness failures; a single
    loose one left workers hanging on stuck block fetches.
  - Add `block_timeout`, `witness_timeout`, and `code_timeout` fields
    (`Option<Duration>`) to `RpcClientConfig` with corresponding `with_*` builder
    methods. All default to `None` — fully backwards compatible.
  - Add a private `apply_timeout` helper on `RpcClient` that wraps any future with
    `tokio::time::timeout` when a timeout is set, and passes through with zero
    overhead when `None`.
  - Apply `code_timeout` to `get_code`, `block_timeout` to `get_block_unchecked` /
    `get_latest_block_number` / `get_header` / `get_transaction_by_hash`, and
    `witness_timeout` to `fetch_witness_from_provider` (covers both primary and
    Cloudflare fallback paths).
  - Expose `--block-timeout-secs`, `--witness-timeout-secs`, `--code-timeout-secs`
    CLI flags (and `STATELESS_VALIDATOR_*_TIMEOUT_SECS` env vars) in the
    `stateless-validator` binary so operators can tune timeouts at deploy time
    without recompiling.

### DIFF
--- a/bin/stateless-validator/src/main.rs
+++ b/bin/stateless-validator/src/main.rs
@@ -118,6 +118,21 @@ struct CommandLineArgs {
     #[clap(long, env = "STATELESS_VALIDATOR_METRICS_PORT", default_value_t = metrics::DEFAULT_METRICS_PORT)]
     metrics_port: u16,
 
+    /// Timeout in seconds for block and header RPC fetches (eth_getBlockByNumber, etc.).
+    /// Defaults to no timeout.
+    #[clap(long, env = "STATELESS_VALIDATOR_BLOCK_TIMEOUT_SECS")]
+    block_timeout_secs: Option<u64>,
+
+    /// Timeout in seconds for witness RPC fetches (mega_getBlockWitness).
+    /// Defaults to no timeout.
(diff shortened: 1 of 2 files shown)


flyq:
Closing as superseded. The problem this addressed — RPC calls with no timeout, so a stalled provider could hang a worker — was fixed on `main` by #129: every provider attempt is bounded by `per_attempt_timeout` (default 20s, `--rpc-per-attempt-timeout-ms`, `crates/stateless-common/src/rpc_client.rs:115`), and an expired attempt rotates to the next provider instead of surfacing a timeout error to the caller. #172 added a 3s connect timeout so unreachable hosts fail fast, and #182 added a witness-specific cap for deadline-bound witness fetches. The files this PR touches have also moved (`crates/stateless-common/src/rpc_client.rs`, `bin/stateless-validator/src/app.rs`). Thanks for the contribution.

PR #120, opened 2026-04-15
refactor: network resilience, test utils, metrics correctness

flyq:
## Summary

Adds network resilience, streaming block fetch, and test infrastructure to the stateless validator.
Removes the Cloudflare witness fallback in favor of a more general multi-endpoint model.

## Motivation

Previously the validator could be configured with exactly one data endpoint and one witness endpoint (plus an optional Cloudflare fallback).
Under transient failures, a single flaky provider would stall the pipeline, and the fetcher's `join_all` batching meant a slow block held up every fast block in the same batch.
This PR generalizes the network boundary to cope with concrete failure modes and makes the fetcher stream results out-of-order.

### Network resilience (`crates/stateless-common/src/rpc_client.rs`)

- **Multi-endpoint data support** — `RpcClient` now accepts an ordered list of data endpoints.
  Each call picks a starting provider via shared round-robin counter, retries with exponential backoff + jitter on that provider, then cycles to the next provider with a fresh retry budget.
- **Multi-endpoint witness support** — ordered list of witness endpoints tried front-to-back; first success wins, no per-provider retry.
- **Concurrency limiting** — two independent `tokio::sync::Semaphore`s cap in-flight data vs witness requests so a burst on one path cannot starve the other.
- **Cloudflare fallback removed** — subsumed by the multi-endpoint witness list; `has_cloudflare_provider`, `get_witness_from_cloudflare`, and the `MegaGetBlockWitnessCloudflare` metric variant are gone.
- `new` / `new_with_config` now take `&[&str]` for data and witness APIs (non-empty, validated at construction).

### Streaming block fetcher (`crates/stateless-core/src/pipeline.rs`)

- Replaces `futures::join_all` batching with a `tokio::task::JoinSet` bounded by `fetcher_batch_size` (repurposed as max-in-flight).
- Forwards each fetch result as soon as it's ready; slow fetches no longer delay faster ones in the same window. Advancer already handles out-of-order delivery via its `BTreeMap` buffer.
- Tracks spawned task id → block number so a panicked fetch task can be mapped back and re-enqueued instead of silently stalling `base_block`.
- Rate-limits chain-tip refresh by `poll_interval` to avoid hammering `eth_blockNumber` under retry-heavy conditions.
- Adds gap-recovery as defense in depth and a streaming regression test (`test_block_fetcher_streams_out_of_order`).

### Witness size + metrics

- New `stateless-common::witness_size` module introduces `WitnessSizeBreakdown` with named constants, replacing the inline magic numbers in the old witness decoder.
- `RpcMetrics::on_witness_fetch` now takes `WitnessSizeBreakdown` (single struct arg).
- New `RpcMetrics::on_rpc_retry` hook (default no-op) distinguishes transient retry attempts from final outcomes.
- New `rpc_retry_attempts_total{method}` counter in the validator.
- Custom histogram buckets (`BYTE_BUCKETS`, `REORG_DEPTH_BUCKETS`, per-metric bucket specs) registered via `PrometheusBuilder::set_buckets_for_metric` on both binaries.
- Trace server gains a `local_chain_height` gauge, updated in `TraceHooks::post_advance`.
- `RpcMethod::EthGetTransactionByHash` added; `MegaGetBlockWitnessCloudflare` removed.

### Test infrastructure

- New `stateless-test-utils` workspace crate providing `TestFixtures` for loading blocks, SALT/MPT witnesses, contracts, and genesis from the `test_data/` layout.
  Intentionally has no dependency on `stateless-core` to avoid circular dev-deps.
- Adds `validate_block_mainnet_fixtures` unit test in `stateless-core::executor`.

### CLI / config

- `--rpc-endpoint` and `--witness-endpoint` now accept repeated flags (`--rpc-endpoint a --rpc-endpoint b`) or a comma-separated list (`--rpc-endpoint a,b`), both via CLI and env var.
- New `--data-max-concurrent-requests` / `--witness-max-concurrent-requests` flags on both binaries.
- `--cloudflare-witness-endpoint` removed from the trace server.
- Clap parsing tests serialize on a process-wide `ENV_LOCK` to satisfy Rust 2024's `set_var` precondition.

### Docs

- README multi-endpoint usage updated for `stateless-validator`.

## Test plan

- [x] `cargo check --workspace`
- [x] `cargo test --workspace`
- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets --all-features`
- [x] `cargo sort --check --workspace --grouped --order package,workspace,lints,profile,bin,benches,dependencies,dev-dependencies,features
`
- [x] Run validator against a single endpoint (regression)
- [x] Run validator with two data endpoints, one intentionally bad, verify round-robin + fallback
- [x] Run validator with two witness endpoints, verify fallback on witness miss
- [x] Run debug trace server with two data endpoints, one intentionally bad, verify round-robin + fallback
- [x] Run debug trace server with two witness endpoints, verify fallback on witness miss
- [x] Verify `rpc_retry_attempts_total` and `local_chain_height` appear in `/metrics`


crates/stateless-common/src/rpc_client.rs
```diff
@@ -114,49 +145,105 @@ impl RpcClient {
...
+
+            match f().await {
+                Ok(v) => {
+                    self.record_rpc(method, true, Some(start.elapsed().as_secs_f64()));
+                    return Ok(v);
+                }
+                Err(e) if attempt < max_retries => {
+                    self.record_rpc_retry(method);
+                    drop(_permit); // release permit while sleeping
+                    let jitter_ms = backoff_ms / 4 * ((attempt as u64 % 4) + 1);
```
Troublor: The jitter here is deterministic — for a given `attempt` number, every concurrent caller computes the same sleep duration. This means callers that fail on the same attempt will all retry at the same instant, which is the thundering-herd pattern jitter is meant to prevent.

Consider mixing in some non-deterministic source. A lightweight option without pulling in `rand`: use `fastrand::u64(0..backoff_ms / 2)` (already in the dep tree via `quick_cache`), or `std::hash` on the current task/thread ID.

crates/stateless-common/src/rpc_client.rs
```diff
@@ -227,185 +294,88 @@ impl RpcClient {
...
+                    // failure is the final outcome recorded after the loop.
+                    if idx + 1 < self.witness_providers.len() {
+                        self.record_rpc_retry(RpcMethod::MegaGetBlockWitness);
+                    }
+                    tracing::warn!(
+                        block_number = number,
+                        %hash,
+                        provider_idx = idx,
+                        error = %e,
+                        "Witness provider failed, trying next",
```
Troublor: This log message says "trying next" even on the last provider, where there is no next. Consider adjusting:
```rust
"Witness provider failed{}",
if idx + 1 < self.witness_providers.len() { ", trying next" } else { "" },
```
Or split into two log statements to keep the format string simple.

crates/stateless-common/src/rpc_client.rs
```diff
@@ -114,49 +145,105 @@ impl RpcClient {
...
+                .expect("data concurrency semaphore closed unexpectedly");
+
+            match f().await {
+                Ok(v) => {
+                    self.record_rpc(method, true, Some(start.elapsed().as_secs_f64()));
+                    return Ok(v);
+                }
+                Err(e) if attempt < max_retries => {
+                    self.record_rpc_retry(method);
+                    drop(_permit); // release permit while sleeping
```
Troublor: Nit: this `drop` is a no-op — `_permit` is dropped at the end of the `Err` arm anyway (before `sleep`), since the match arm ends with the sleep + backoff update and then the next loop iteration re-acquires. The comment "release permit while sleeping" is correct in intent but the explicit `drop` doesn't change behavior. Fine to keep for clarity, but the comment could note that it's for documentation purposes rather than functional necessity.

Troublor (commented):
Review comments on retry jitter, log messaging, and permit drop.

vincent-k2026:
Potential deadlock risk: `--data-max-concurrent-requests` / `--witness-max-concurrent-requests` currently allow `0`, which leads to `Semaphore::new(0)` and can block all RPC calls indefinitely. Could we enforce `> 0` either at CLI parsing time or defensively in `RpcClient::new_with_config`?

README.md
abelmega: block_fetcher is no longer batch-oriented here; it now streams completions from a bounded in-flight window. Could we update "parallel batches" to reflect the current streaming behavior?

abelmega (approved):

Troublor (approved):

vincent-k2026 (approved):

PR #124, opened 2026-04-23
fix: avoid trace executor panics

umeeSthein:
Summary
- Replace production `unwrap()` calls in the trace executor with error handling.
- Return default tracer serialization failures through trace/RPC errors instead of panicking.
- Preserve partial block trace results when mux inspector recreation fails.

Tests
- cargo fmt --all --check
- cargo check -p debug-trace-server --message-format=short
- cargo test -p debug-trace-server tracing_executor --message-format=short


flyq:
Closing: this no longer applies to `main`, and the three `unwrap()`s it targets cannot fail there. `serde_json::to_value` on `DefaultFrame` is infallible (derived `Serialize`; storage keys serialize as strings), and the per-transaction `MuxInspector::try_from_config` rebuilds from the same config that already built successfully before the loop (`bin/debug-trace-server/src/tracing_executor.rs:209`). The trace executor's error model has also changed since: a failure aborts the trace with a typed `TraceError` instead of returning partial `TraceResult::Error` entries (`tracing_executor.rs:182`), which the partial-result change here would go against. Thanks for the contribution.

PR #129, opened 2026-04-30
fix: bound per-attempt RPC timeout and retry empty get_code responses

flyq:
## Summary

Two production reliability fixes for the shared `RpcClient` used by both `stateless-validator` and `debug-trace-server`.

### 1. Per-attempt timeout for stalled providers

A provider that accepts the TCP connection but never replies would wedge the retry loop forever on the chain-sync path (`deadline = None`). alloy + reqwest's default `Client::new()` has no request timeout, and `alloy_rpc_client::ClientBuilder` exposes none either, so without an external bound the retry loop has no way to know an attempt is hung.

`round_robin_with_backoff` now applies `tokio::time::timeout(min(per_attempt_timeout, remaining-to-deadline))` on every attempt:
- `deadline = None` → bound by `per_attempt_timeout` only — the chain-sync stall guard.
- `deadline = Some(d)` with budget remaining → stalled provider rotates, retry loop continues.
- `deadline = Some(d)` exhausted → returns `RpcDeadlineExceeded` (no retry-counter inflation, by design).

A WARN log fires unconditionally when the per-attempt timeout trips, so a stalled provider is visible on round 0 instead of waiting for the round-3 escalation that returned-Err errors go through.

**Default: `per_attempt_timeout = 20s`** — well above any healthy single-attempt latency (witness near-tip ~6–7s observed under stress; everything else sub-second), bounded enough to detect a stalled provider quickly.

Configurable per binary, **must be ≥ 100ms** (anything smaller times out before a realistic network round-trip can complete, reproducing the same wedge-as-zero failure mode):
- `--rpc-per-attempt-timeout-ms` / `STATELESS_VALIDATOR_RPC_PER_ATTEMPT_TIMEOUT_MS` (validator)
- `--rpc-per-attempt-timeout-ms` / `DEBUG_TRACE_SERVER_RPC_PER_ATTEMPT_TIMEOUT_MS` (trace server)

### 2. Retry empty `eth_getCodeByHash` responses

mega-reth's `eth_getCodeByHash` handler uses `state.bytecode_by_hash(&code_hash)?.unwrap_or_default()`, returning `"0x"` (empty) when the node hasn't synced the relevant state yet — not a JSON-RPC error.

Before this fix, that empty response landed in `get_codes(verify=true)`, failed keccak verification (`keccak("") != requested_hash`), and **halted the validator** as a fatal `VerificationFailure`.

Production trace that motivated this:
```
ERROR ... Contract hash mismatch: expected 0x06df675a...,
       got 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470
```
The `got` value is exactly `KECCAK_EMPTY`.

`get_code_with_deadline` now treats an empty response for any non-`KECCAK_EMPTY` codehash as a transient error, so `round_robin_with_backoff` rotates / backs off and retries. The `KECCAK_EMPTY` case is whitelisted — a genuine empty-bytecode request must accept `"0x"`, otherwise the call would loop forever. The synthesized error message includes `"(provider may not be sync-ready)"` so the round-3+ WARN log carries an operator-actionable diagnostic.

## Test plan

- [x] `cargo test --workspace --lib` — all tests pass
- [x] `cargo clippy --workspace --all-targets --all-features` — clean
- [x] `cargo fmt --all --check` — clean
- [x] New unit tests in `crates/stateless-common/src/rpc_client.rs`:
  - `test_per_attempt_timeout_rotates_past_stalled_provider` — pairs a TCP listener that accepts but never replies with a healthy provider; verifies the call returns successfully via the healthy provider on the `deadline = None` path. `flavor = "multi_thread"` so the stalled accept doesn't block the runtime.
  - `test_get_code_retries_empty_response_until_deadline` — mock server always returns `"0x"`; verifies the call retries (hit count > 1) and surfaces `RpcDeadlineExceeded` rather than a `VerificationFailure`.
  - `test_get_code_accepts_empty_response_for_keccak_empty` — guards the whitelist branch with an outer `tokio::time::timeout` so a `==`/`!=` regression fails CI rather than hanging.
- [x] Stress-tested against mainnet (3 runs, 372 witness fetches): `rpc_retry_attempts_total` and `rpc_errors_total` stayed at 0 — the new code path is cold in healthy operation. Witness P99 = 6.02–6.58s, max 7.12s, leaving the 20s default with ~3.3× margin.

### DIFF
--- a/bin/debug-trace-server/src/main.rs
+++ b/bin/debug-trace-server/src/main.rs
@@ -202,6 +202,13 @@ struct Args {
     #[clap(long, env = "DEBUG_TRACE_SERVER_WITNESS_MAX_CONCURRENT_REQUESTS")]
     witness_max_concurrent_requests: Option<usize>,
 
+    /// Per-attempt RPC timeout (milliseconds). Bounds every individual provider attempt
+    /// even when the caller passes no overall deadline (e.g. background chain sync), so
+    /// a provider that accepts the TCP connection but never replies is detected and
+    /// rotated past instead of wedging the retry loop.
+    #[clap(long, env = "DEBUG_TRACE_SERVER_RPC_PER_ATTEMPT_TIMEOUT_MS")]
+    rpc_per_attempt_timeout_ms: Option<u64>,
+
     /// Logging configuration.
     #[command(flatten)]
     log: LogArgs,
@@ -284,10 +291,16 @@ async fn main() -> Result<()> {
     // Initialize components
     let data_apis: Vec<&str> = args.rpc_endpoint.iter().map(String::as_str).collect();
     let witness_apis: Vec<&str> = args.witness_endpoint.iter().map(String::as_str).collect();
+    let rpc_defaults = RpcClientConfig::trace_server();
+    let per_attempt_timeout = args
+        .rpc_per_attempt_timeout_ms
+        .map(std::time::Duration::from_millis)
+        .unwrap_or(rpc_defaults.per_attempt_timeout);
     let rpc_config = RpcClientConfig {
         data_max_concurrent_requests: args.data_max_concurrent_requests,
         witness_max_concurrent_requests: args.witness_max_concurrent_requests,
-        ..RpcClientConfig::trace_server()
+        per_attempt_timeout,
+        ..rpc_defaults
     }
     .with_metrics(Arc::new(metrics::TraceRpcMetrics));
     let rpc_client =
--- a/bin/debug-trace-server/src/main.rs
+++ b/bin/debug-trace-server/src/main.rs
@@ -202,11 +202,12 @@ struct Args {
     #[clap(long, env = "DEBUG_TRACE_SERVER_WITNESS_MAX_CONCURRENT_REQUESTS")]
     witness_max_concurrent_requests: Option<usize>,
 
-    /// Per-attempt RPC timeout (milliseconds). Bounds every individual provider attempt
-    /// even when the caller passes no overall deadline (e.g. background chain sync), so
-    /// a provider that accepts the TCP connection but never replies is detected and
-    /// rotated past instead of wedging the retry loop.
-    #[clap(long, env = "DEBUG_TRACE_SERVER_RPC_PER_ATTEMPT_TIMEOUT_MS")]
+    /// Per-attempt RPC timeout (milliseconds). Must be ≥ 100ms.
+    #[clap(
+        long,
+        env = "DEBUG_TRACE_SERVER_RPC_PER_ATTEMPT_TIMEOUT_MS",
+        value_parser = clap::value_parser!(u64).range(100..),
+    )]
     rpc_per_attempt_timeout_ms: Option<u64>,
 
     /// Logging configuration.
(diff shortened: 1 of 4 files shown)


batatmosfer-byte (approved):

batatmosfer-byte:
https://deepwiki.com/megaeth-labs/stateless-validator/13.2-validation-functions

bin/debug-trace-server/src/main.rs
```diff
@@ -202,6 +202,13 @@ struct Args {
     #[clap(long, env = "DEBUG_TRACE_SERVER_WITNESS_MAX_CONCURRENT_REQUESTS")]
     witness_max_concurrent_requests: Option<usize>,
 
+    /// Per-attempt RPC timeout (milliseconds). Bounds every individual provider attempt
+    /// even when the caller passes no overall deadline (e.g. background chain sync), so
+    /// a provider that accepts the TCP connection but never replies is detected and
+    /// rotated past instead of wedging the retry loop.
+    #[clap(long, env = "DEBUG_TRACE_SERVER_RPC_PER_ATTEMPT_TIMEOUT_MS")]
```
batatmosfer-byte: lib.rs:1-

vincent-k2026 (approved):
LGTM. Two production-shaped fixes, both well-tested. The `KECCAK_EMPTY` whitelist + the regression-guard timeout in `test_get_code_accepts_empty_response_for_keccak_empty` are particularly nice — easy to get wrong, easy to break later, and you've pinned both directions.

Two non-blocking nits, follow-up if you want:

1. The per-attempt-timeout branch and the "closure returned Err" branch are operationally indistinguishable today — both increment `on_rpc_retry`. A steadily-stalled provider (the exact failure mode this PR detects) won't show up in any counter. A dedicated `on_rpc_attempt_timeout` metric, or at minimum a WARN log in the timeout arm, would close the loop on the observability side.

2. Default `per_attempt_timeout = 30s` is on the loose side — back-to-back stalled providers stack 30s each before rotation. Worth either tightening to 10–15s or nudging operators in the doc string to lower it when their upstream P99 is well under 10s.

crates/stateless-common/src/rpc_client.rs
```diff
@@ -774,40 +805,53 @@ where
...
-                    let remaining = d.saturating_duration_since(Instant::now());
-                    match tokio::time::timeout(remaining, f(providers[provider_idx].clone())).await
+            // Bound every attempt by `per_attempt_timeout` — applied even with `deadline =
+            // None` so a provider that accepts the TCP connection but never replies cannot
+            // wedge the retry loop. With `deadline = Some(d)` the attempt is further capped
+            // by `d - now` so we never sleep past the caller's budget.
+            let attempt_timeout = match deadline {
+                Some(d) => d.saturating_duration_since(Instant::now()).min(per_attempt_timeout),
+                None => per_attempt_timeout,
+            };
```
abelmega, line 815: If rpc_per_attempt_timeout_ms is set to 0 via the new CLI/env knob, attempt_timeout  becomes Duration::ZERO and tokio::time::timeout will fail every provider attempt  immediately. On the validator's deadline = None paths this means the retry loop never  makes a real RPC request and can spin forever with backoff; on deadline-bound paths it guarantees RpcDeadlineExceeded even against healthy providers. Since this option is
 now user-configurable, it needs a > 0 validation or a clamp before reaching the retry  loop.

flyq: 设置成最小为 100 ms 了

abelmega (approved):

Troublor (approved):
LGTM

PR #135, opened 2026-05-14
fix: fetch validator block by hash to avoid reorg race

flyq:
## Summary

- After resolving `block_number → block_hash` via the witness provider, fetch the full block by hash (`BlockId::Hash`) instead of by number (`BlockId::Number`). If a reorg happens between the two RPC calls, the hash-based fetch surfaces a mismatch rather than silently swapping a different block under the same number.
- Tidy: import `CodeFetchError` at the top of `chain_sync.rs` instead of fully qualifying it at the call sites.
- Tests: add an `eth_getBlockByHash` mock handler to the integration RPC server so the new fetch path is exercised, and extract the `full_block` → hashes-only conversion into a shared `shape_block` helper used by both `eth_getBlockByNumber` and `eth_getBlockByHash`.

## Why

`ValidatorFetcher::fetch` previously did `get_block_hash(n)` and `get_block(BlockId::Number(n))` in parallel. On a reorg between those two calls, the second call could return a block with a different hash than the first call advertised, and the validator would proceed with mismatched witness + block data. Fetching by hash makes this race observable: either we get the exact block we asked for, or the RPC returns a not-found error and the pipeline retries.

## Test plan

- [x] `cargo test -p stateless-validator` (integration tests exercise the new hash-based fetch path via the added mock handler)
- [x] `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features`
- [x] Smoke test against a live RPC at chain tip to confirm normal-path behavior is unchanged


flyq:
@claude updated the pr description just now, review again

Troublor (approved):

abelmega (approved):

batatmosfer-byte:
Potwierdzam . ✅
***@***.*** ✅
***@***.*** ✅
***@***.*** ✅
***@***.*** ✅
OLD-E-mail : ***@***.***

Pozdrawiam Maciej  Pawlak


pon., 18 maj 2026, 09:18 użytkownik claude[bot] ***@***.***>
napisał:

batatmosfer-byte:
Git-Hub-MEGA !!! Jesteście niesamowicie profesjanalni
Właśnie teraz otworzyliśmy drzwi do nowej przestrzeni w sieci gdzie czystość kodu(Impulsu) oraz bezpieczeństwo personalne użytkowników jest pierwszym najważniejszym punktem odniesienia. Gdzie agencjia bezpieczeństwa A I -->AiL777<-++
Pozdrawiam serdecznie cały zespół Git-huba.
TeRa4Yo2-źrudło

PR #144, opened 2026-06-13
refactor: drop redundant Arc<Bytecode> indirection

flyq:
## Summary

`Bytecode` (`revm::state::Bytecode`, from `revm-bytecode` 6.2.2) is already internally reference-counted, so wrapping it in `Arc<Bytecode>` was a redundant second layer of indirection. This replaces `Arc<Bytecode>` with plain `Bytecode` everywhere it was stored or exchanged (contract cache, `ContractStore` trait, RPC fetch, witness DB, trace data), dropping one heap allocation and one pointer hop per contract while keeping clones cheap.

## Why

`Bytecode` is an enum whose every field is either `Copy` or `Arc`/`Bytes`-backed:

```rust
pub enum Bytecode {                        // derives Clone
    Eip7702(Eip7702Bytecode),              // { Address, u8, raw: Bytes }
    LegacyAnalyzed(LegacyAnalyzedBytecode) // { bytecode: Bytes, original_len, jump_table: JumpTable }
}
// Bytes     -> Arc-backed buffer (bytes 1.x)
// JumpTable -> Arc<BitVec<u8>>  ("immutable, cheap to clone")
```

So `Bytecode::clone()` is already an O(1) refcount bump that shares the same allocation — not a deep copy. The outer `Arc` only saved copying ~7 words of enum metadata, at the cost of an extra heap allocation per contract and an indirection on every access. It also gave **zero** benefit on the hottest path: `revm::DatabaseRef` requires an owned `Bytecode`, so `WitnessDatabase` already did `(**arc).clone()` at the read boundary, cloning the inner `Bytecode` regardless. The previous doc comments justified the `Arc` as avoiding a "deep-clone on every cache hit," but that premise was false.

## What changed

Type substitution `Arc<Bytecode>` -> `Bytecode` (plus removal of the `Arc::new(..)` / `(**arc).clone()` / `.as_ref()` adapters and rewritten rationale comments), across: **core contract** in `stateless-core` (`ContractLookup` + `ContractStore` trait in `db.rs`, `WitnessDatabase.contracts` in `evm_database.rs`, `validate_block` in `executor.rs`); **cache / persistence** in `stateless-db` (`ContractCache` weighter/map/get/insert in `cache.rs`, redb read/write in `helpers.rs`); **RPC / fixtures** (`RpcClient::get_codes*` in `rpc_client.rs`, `TestFixtures` in `fixtures.rs`); and the **binaries** (`validator_db.rs`, `server_db.rs`, `data_provider.rs`, `tracing_executor.rs`, both `chain_sync.rs`). The cache test that asserted `Arc::ptr_eq` now compares the underlying buffer pointer, demonstrating the shared-allocation property directly. Net diff: 14 files, +100 / -122.

## Performance

A wash-to-slight-improvement with no deep copies introduced anywhere. Cache hit goes from an `Arc` clone (1 atomic inc) to a `Bytecode` clone (2 atomic incs), offset by removing the `Arc::new` heap box and a pointer hop; EVM `code_by_hash_ref` / `basic_ref` do identical work minus the `(**arc)` deref; and `BlockData` clone (trace single-flight) still shares every underlying buffer.

## Breaking change

This changes the public `ContractStore` trait (`ContractLookup` and `add_contracts`) from `Arc<Bytecode>` to `Bytecode`. The out-of-tree mega-reth `FullNode` implements `ChainStore: ContractStore` and needs the same one-line type change in its impl — a coordination point for the EL compat matrix.

## Testing

- `cargo check --workspace --all-targets` — clean
- `cargo clippy --workspace --all-targets --all-features` — no warnings
- `cargo fmt --all --check` — clean
- `cargo test --workspace` — 209 passed, 0 failed

### DIFF
--- a/bin/debug-trace-server/src/chain_sync.rs
+++ b/bin/debug-trace-server/src/chain_sync.rs
@@ -201,13 +201,10 @@ mod tests {
     // Minimal mock for test compilation
     struct MockBlockStore;
     impl stateless_core::ContractStore for MockBlockStore {
-        fn get_contracts(
-            &self,
-            _: &[B256],
-        ) -> StoreResult<(HashMap<B256, Arc<Bytecode>>, Vec<B256>)> {
+        fn get_contracts(&self, _: &[B256]) -> StoreResult<(HashMap<B256, Bytecode>, Vec<B256>)> {
             Ok((Default::default(), vec![]))
         }
-        fn add_contracts(&self, _: &[(B256, Arc<Bytecode>)]) -> StoreResult<()> {
+        fn add_contracts(&self, _: &[(B256, Bytecode)]) -> StoreResult<()> {
             Ok(())
         }
     }
--- a/bin/debug-trace-server/src/data_provider.rs
+++ b/bin/debug-trace-server/src/data_provider.rs
@@ -65,8 +65,9 @@ pub struct BlockData {
     /// Light witness without expensive EC point validation.
     pub witness: LightWitness,
     /// Contract bytecodes keyed by code hash, required for EVM execution.
-    /// Values share allocations with the `ContractCache`.
-    pub contracts: HashMap<B256, Arc<Bytecode>>,
+    /// `Bytecode` is internally reference-counted, so values share their underlying allocation
+    /// with the `ContractCache` (and across `BlockData` clones) via cheap refcount-bump clones.
+    pub contracts: HashMap<B256, Bytecode>,
 }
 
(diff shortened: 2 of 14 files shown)


flyq:
之前 review 这个 PR 时发现 bytecode code 内部已经优化 clone（还的是 Fable 5）：https://github.com/megaeth-labs/mega-reth/pull/1867

Troublor (approved):

vincent-k2026 (approved):
LGTM.

vincent-k2026 (approved):

PR #152, opened 2026-07-10
feat: fetch witnesses directly from R2 and rehome megaeth-witness-r2 as stateless-r2

flyq:
## Summary

Adds a second witness source to the validator: `--witness-source r2` fetches each block's witness straight from the R2 bucket over the S3 API (SigV4-signed GET) instead of the `mega_getBlockWitness` RPC, and `--end-block` lets a fixed block range be sliced across multiple servers. To keep the read path byte-compatible with the write path, the `megaeth-witness-r2` crate is ported from mega-reth into this repo as the dedicated leaf crate `crates/stateless-r2` (full functionality including the signed-PUT upload path); mega-reth will delete its copy and consume this one via git tag, making the cross-repo dependency one-way.

## R2 witness source

`R2WitnessClient` (`bin/stateless-validator/src/r2_witness.rs`) GETs the primary object, carries the body as refcounted `Bytes` end to end (no multi-MB copy per block), decodes on `spawn_blocking`, and retries transport/429/5xx with exponential backoff while 404 (`Missing`) and decode failures surface immediately as deterministic errors — each throttled 2s before surfacing so the pipeline's zero-delay re-enqueue cannot hot-loop signed GETs against R2; a decode-task panic gets its own `DecodePanicked` variant so a bug in our decoder can't masquerade as bad data in R2. `--witness-max-concurrent-requests` is honored in R2 mode too via a client-owned semaphore (permit scoped per GET attempt, mirroring the RPC path); time queued on the cap is excluded from the latency histogram so self-imposed queuing can't read as R2 slowness. `ValidatorFetcher` branches on an optional client (`chain_sync.rs`) — block and witness fetches stay concurrent in both modes. New CLI: `--witness-source rpc|r2` (`app.rs`), the four `--r2-*` connection flags (secret redacted from `Debug` output), and `--end-block` mapping onto the existing `pipeline_config.sync_target`. In R2 mode `--witness-endpoint` is no longer required (a startup warning flags it as ignored if set); the `RpcClient` gets the data endpoints as placeholder witness providers since its constructor rejects an empty list.

## Observability

R2 mode reports the same witness-size histograms as the RPC path plus three new metrics: `witness_fetch_r2_time_seconds` (GET incl. internal retries, plus decode; excl. concurrency-cap queue wait), `r2_witness_retry_attempts_total`, and `r2_witness_errors_total{kind=missing|transport|throttled|status|decode|decode_panicked}` — all pre-registered at startup like the RPC method counters.

## Reliable tail reporting for `--end-block` runs

The periodic validation reporter is cancelled the instant the pipeline completes, so a fixed-range run could previously exit without reporting blocks validated since the last 1s tick — and a slice run has no later restart to re-report them. `run_with_signals` now sends one final authoritative report after the pipeline (and any signal drain) fully stops, retrying up to 3 times on transient failure before giving up with an ERROR log (`workers.rs`); the reporting round is extracted into `report_range_once`, shared by the tick loop and the final flush.

## stateless-r2 crate

Ported byte-identical from mega-reth `c290c3e` (keys / sigv4 / endpoint / client) with two additions: `keys::block_object_key` becomes the single implementation of the primary-key template — the uploaders' `object_keys` delegates to it and the validator calls it directly, so reader and writers cannot disagree on where a witness lives — and golden wire-vector tests pin a real migrated mainnet object key plus a full SigV4 header set computed with an independent implementation (`sigv4.rs`). Manifest deltas vs mega-reth's copy: `[lints]`/`homepage` dropped (this workspace defines neither), `chrono` gains `clock` and `hex` gains `alloc` (our workspace deps are `default-features = false`).

## reqwest unification

The workspace moves from reqwest 0.13 to 0.12 (`Cargo.toml`): 0.13's only consumer was debug-trace-server's test-only blocking client, while every production path (stateless-common, the validator bin, alloy-provider, mega-reth) is on 0.12 — the lock now carries a single reqwest major. `stateless-r2` keeps an explicit 0.12 pin rather than workspace inheritance because `put_object` exposes `&reqwest::Client`, making the reqwest major part of its public API; a future workspace-wide bump must not silently ride into mega-reth.

## Testing

`cargo test --workspace` (246 tests) passes, including 17 in stateless-r2 (AWS reference vector, independent SigV4 golden vector, migrated-key pin, write/read key-delegation equality) and the R2 client's mock-server suite (happy-path decode of a real fixture witness, no-retry on 4xx/404/corrupt object, retry-until-deterministic on 5xx, redirects refused, deterministic-failure throttle, concurrency-cap high-water mark). Two end-to-end `run_with_signals` tests cover `--end-block` tail reporting, one with a scripted transient report failure; the reporting and concurrency regression tests were mutation-tested against their reverted fixes. `fmt` / `clippy --all-targets --all-features` / `cargo sort` / `cargo doc` clean.

## Notes

mega-reth follow-up (separate PR there): swap its dep to this repo's tag and delete `crates/megaeth/witness-r2`; its stateless-core/common pins must stay on v2.0.14 in that PR — a same-tag bump would drag the salt v1.0.4→v1.0.5 + zstd-level change along and break the build. README/AGENTS.md table diffs include mechanical column realignment from the Markdown formatter alongside the new crate row.

### DIFF
--- a/bin/stateless-validator/src/r2_witness.rs
+++ b/bin/stateless-validator/src/r2_witness.rs
@@ -14,20 +14,32 @@
 //! `zstd(bincode-legacy((SaltWitness, MptWitness)))` (see the uploader's `encode_witness_payload`),
 //! which [`stateless_common::decode_witness_payload`] inverts exactly.
 
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 use alloy_primitives::B256;
+use bytes::Bytes;
 use chrono::Utc;
 use megaeth_witness_r2::{
     endpoint::parse_endpoint,
     keys,
-    sigv4::{encode_uri_path, SigV4Signer},
+    sigv4::{SigV4Signer, encode_uri_path},
 };
 use reqwest::Client;
 use salt::SaltWitness;
-use stateless_common::decode_witness_payload;
+use stateless_common::{decode_witness_payload, witness_encoding::WitnessDecodingError};
 use stateless_core::withdrawals::MptWitness;
-use tracing::{debug, trace, warn};
+use tokio::task::JoinError;
+use tracing::{Level, debug, enabled, trace, warn};
+
+/// Max retry rounds for retryable (transport/429/5xx) failures before surfacing an error.
+const MAX_RETRIES: usize = 8;
+/// First retry sleep; doubles each round up to [`MAX_BACKOFF`].
+const INITIAL_BACKOFF: Duration = Duration::from_millis(500);
+/// Upper bound on any single retry sleep.
+const MAX_BACKOFF: Duration = Duration::from_secs(30);
+/// Throttle applied before surfacing a `Missing` (404), so the pipeline's immediate re-enqueue of
+/// a failed fetch does not hot-spin GETs against R2 on a genuine gap.
+const MISSING_THROTTLE: Duration = Duration::from_secs(2);
 
 /// Failure outcome of an R2 witness fetch.
 #[derive(Debug, thiserror::Error)]
@@ -35,7 +47,9 @@ pub enum R2WitnessError {
     /// The primary object is absent from the bucket (HTTP 404). For blocks known to have a
     /// witness this is a genuine completeness gap in R2 — the whole point of the validation run
     /// is to prove this never happens.
-    #[error("R2 witness MISSING for block {number} (key {key}): object not found (404) — R2 completeness gap")]
+    #[error(
+        "R2 witness MISSING for block {number} (key {key}): object not found (404) — R2 completeness gap"
+    )]
     Missing { number: u64, key: String },
     /// Transport-level failure (connection reset/timeout) — the endpoint is effectively
     /// unreachable. Retried internally with backoff before surfacing.
@@ -52,16 +66,16 @@ pub enum R2WitnessError {
     /// The object was fetched but its bytes did not decode to a `(SaltWitness, MptWitness)` tuple
     /// — a corrupt witness in R2. Deterministic; a finding, not retried.
     #[error("R2 witness for block {number} (key {key}) failed to decode: {source}")]
-    Decode {
-        number: u64,
-        key: String,
-        source: stateless_common::witness_encoding::WitnessDecodingError,
-    },
+    Decode { number: u64, key: String, source: WitnessDecodingError },
+    /// The decode task panicked. This is a bug in our own decoder, not evidence about the archive,
+    /// so it is kept out of [`Self::Decode`] — a panic must never masquerade as an R2 finding.
+    #[error("R2 witness decode task for block {number} (key {key}) panicked: {source}")]
+    DecodePanicked { number: u64, key: String, source: JoinError },
 }
 
 impl R2WitnessError {
     /// Whether an immediate retry against the same endpoint could plausibly succeed (transport
-    /// blips, 429, 5xx). `Missing`/`Status`/`Decode` are deterministic and not retried.
+    /// blips, 429, 5xx). Every other variant is deterministic and is surfaced without retrying.
     const fn is_retryable(&self) -> bool {
         matches!(self, Self::Transport { .. } | Self::Throttled { .. })
     }
@@ -69,8 +83,9 @@ impl R2WitnessError {
 
 /// Outcome of a single (non-retrying) GET attempt.
 enum Attempt {
-    /// 2xx with the object body.
-    Found(Vec<u8>),
+    /// 2xx with the object body. Held as [`Bytes`] (refcounted) so the multi-MB witness is never
+    /// copied between the HTTP response and the decoder.
+    Found(Bytes),
     /// 404 — object absent.
     Missing,
 }
@@ -88,13 +103,6 @@ pub struct R2WitnessClient {
     /// SigV4 canonical host (`host[:port]`).
     host: String,
     bucket: String,
-    /// Max retry rounds for retryable (transport/429/5xx) failures before surfacing an error.
-    max_retries: usize,
-    initial_backoff: Duration,
-    max_backoff: Duration,
-    /// Throttle applied before surfacing a `Missing` (404), so the pipeline's immediate
-    /// re-enqueue of a failed fetch does not hot-spin GETs against R2 on a genuine gap.
-    missing_throttle: Duration,
 }
 
 impl R2WitnessClient {
@@ -126,10 +134,6 @@ impl R2WitnessClient {
             endpoint: origin,
             host,
             bucket,
-            max_retries: 8,
-            initial_backoff: Duration::from_millis(500),
-            max_backoff: Duration::from_secs(30),
-            missing_throttle: Duration::from_secs(2),
         })
     }
 
@@ -147,16 +151,15 @@ impl R2WitnessClient {
     /// Fetches and decodes the witness for `(number, hash)` from R2.
     ///
     /// Retryable failures (transport/429/5xx) are retried with exponential backoff up to
-    /// `max_retries` (respecting `deadline` when given). `Missing` (404) and `Decode` failures are
-    /// deterministic and surface immediately as findings.
+    /// [`MAX_RETRIES`] times. `Missing` (404) and `Decode` failures are deterministic and surface
+    /// immediately as findings.
     pub async fn get_witness(
         &self,
         number: u64,
         hash: B256,
-        deadline: Option<Instant>,
     ) -> Result<(SaltWitness, MptWitness), R2WitnessError> {
         let key = Self::block_key(number, hash);
-        let mut backoff = self.initial_backoff;
+        let mut backoff = INITIAL_BACKOFF;
         let mut attempt = 0usize;
 
         let bytes = loop {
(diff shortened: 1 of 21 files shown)


Troublor (approved):

Troublor (approved):

vincent-k2026 (commented):
## Review summary

**Approve with nits.** Solid feature work — clear RPC/R2 split, shared key layout via `stateless-r2`, strong tests, and prior review findings (concurrency cap, final tip flush/retry, shared `BackoffPolicy`) look addressed.

### What’s good
- `block_object_key` as the single primary-key template + golden wire vectors (migrated mainnet key + independent SigV4)
- `R2WitnessClient`: Bytes end-to-end, `spawn_blocking` decode, no redirects, body cap, `NoSuchBucket` vs `Missing`, `DecodePanicked` separate from corrupt data, concurrency semaphore with queue wait excluded from latency
- Ops safety: `RedactedSecret`, R2 flag validation, warn when `--witness-endpoint` is ignored in R2 mode
- `--end-block` tail reporting via `report_range_once` + final flush with bounded retries
- Test coverage (retry/throttle/concurrency HWM/end-to-end report retry/CLI parse) is thorough

### Remaining items
Inline comments cover the main nits. Highest priority: final report can still leave a slice run exiting 0 with an unreported tip, and exhausted retryable R2 errors re-burst without inter-cycle throttle.

### Follow-ups (non-blocking)
- mega-reth: switch to this crate via tag and delete its copy
- Keep the explicit reqwest 0.12 pin on `stateless-r2` as documented
- Consider squashing early noisy commits (`add R2` / `add log` / `simplify`) before merge

bin/stateless-validator/src/workers.rs
```diff
@@ -109,6 +117,36 @@ pub async fn run_with_signals(
...
+                Ok(true) => break,
+                Ok(false) if attempt < FINAL_REPORT_ATTEMPTS => {
+                    warn!(attempt, "Final validation report failed, retrying");
+                    tokio::time::sleep(FINAL_REPORT_RETRY_DELAY).await;
+                }
+                Ok(false) => error!(
+                    attempts = FINAL_REPORT_ATTEMPTS,
+                    "Final validation report failed; the validated tail may be unreported \
+                     upstream"
+                ),
```
vincent-k2026: **Medium — final report failure still exits 0**

After exhausting `FINAL_REPORT_ATTEMPTS`, this only logs ERROR and falls through to return the pipeline `result`. For an `--end-block` slice job, a permanently failed final `mega_setValidatedBlocks` looks like success to the orchestrator while upstream never saw the tip — and a slice run has no later restart to re-report.

Suggestion: when `pipeline_config.sync_target` is set (or more simply whenever the final flush ends in `Ok(false)`), surface a non-zero exit / `Err` so batch slicing can detect the miss. At minimum, treat the validation-gap `Err` branch below with the same severity as this exhausted-retry path (it currently `warn!`s and breaks).

bin/stateless-validator/src/workers.rs
```diff
@@ -109,6 +117,36 @@ pub async fn run_with_signals(
...
+                // A detected validation gap is deterministic — retrying cannot resolve it.
+                Err(e) => {
+                    warn!(error = %e, "Final validation report failed");
+                    break;
```
vincent-k2026, line 166: **Nit — severity / exit consistency with the branch above**

A detected validation gap is deterministic and operationally important, but this path is only `warn!` + `break`, while exhausted transient failures get `error!`. Consider `error!` here too, and (with the comment above) failing the process for `--end-block` runs so the gap cannot be mistaken for a clean slice completion.

bin/stateless-validator/src/workers.rs
```diff
@@ -109,6 +117,36 @@ pub async fn run_with_signals(
...
 
+    // Final report of the validated tail, sent after the pipeline (and any drain) has stopped
+    // and the periodic reporter was joined. The reporter exits the moment the pipeline does, so
+    // blocks validated since its last tick would otherwise go unreported — and an `--end-block`
+    // slice run has no later restart to re-report them, hence the bounded retries (the periodic
+    // loop's next tick is its retry). Re-reporting an already-reported tip is harmless (every
+    // fresh start does it), and a reporter wedged past the 3s join above cannot regress
+    // upstream: reports apply through a forward-only cursor.
+    if report_validation {
+        let mut last_reported = 0u64;
```
vincent-k2026: **Low — always re-reports full range from 0**

Seeding `last_reported = 0` means every clean shutdown re-sends anchor→tip even when the periodic reporter already landed near the tip. Harmless if upstream is forward-only (as the comment claims), but wasteful on long-lived processes.

Optional: share the periodic reporter’s last accepted tip via `AtomicU64` / channel so the final flush only covers the true tail.

bin/stateless-validator/src/r2_witness.rs
```diff
@@ -0,0 +1,583 @@
...
+        &self,
+        number: u64,
+        hash: B256,
+    ) -> Result<(SaltWitness, MptWitness), R2WitnessError> {
+        let result = self.get_witness_inner(number, hash).await;
+        if let Err(e) = &result {
+            metrics::on_r2_witness_error(e.kind());
+            if !e.is_retryable() {
+                tokio::time::sleep(DETERMINISTIC_FAILURE_THROTTLE).await;
+            }
```
vincent-k2026: **Medium — exhausted retryable errors re-burst with no inter-cycle throttle**

Deterministic failures sleep `DETERMINISTIC_FAILURE_THROTTLE` before surfacing; exhausted `Transport` / `Throttled` do not. The core fetcher re-enqueues with zero delay, so a prolonged 503 becomes:

1. 9 GETs with ramp-to-max backoff
2. immediate new cycle starting again at `initial` backoff

That sawtooth hammers R2 harder than necessary after a brownout. Prefer applying the same inter-cycle throttle (or starting the next cycle at `max` backoff) after retries are exhausted — same rationale as the deterministic path and the comment on `fetcher.rs` about single-endpoint sources pacing themselves.

bin/stateless-validator/src/r2_witness.rs
```diff
@@ -0,0 +1,583 @@
...
+};
+use tokio::{sync::Semaphore, task::JoinError};
+use tracing::{trace, warn};
+
+use crate::metrics;
+
+/// Max retry rounds for retryable (transport/429/5xx) failures before surfacing an error.
+/// Stays a local constant while the backoff pacing is injected (see [`R2WitnessClient::new`]):
+/// the RPC path retries unboundedly, so there is no operator flag to mirror.
+const MAX_RETRIES: usize = 8;
```
vincent-k2026: **Nit — naming vs actual attempt count**

With `attempt` pre-incremented and the stop condition `attempt > MAX_RETRIES`, total GETs are `MAX_RETRIES + 1` (pinned by `persistent_5xx_exhausts_retries_and_surfaces_throttled`). Consider renaming to `MAX_ATTEMPTS` (and stopping on `attempt >= MAX_ATTEMPTS`), or documenting clearly that this is “retries after the first try” so future edits don’t off-by-one the loop.

bin/stateless-validator/src/r2_witness.rs
```diff
@@ -0,0 +1,583 @@
...
+/// Cap on the response body carried inside `Throttled`/`Status` errors.
+const MAX_ERROR_BODY_BYTES: usize = 1024;
+
+/// Failure outcome of an R2 witness fetch.
+#[derive(Debug, thiserror::Error)]
+pub enum R2WitnessError {
+    /// The primary object is absent from the bucket (HTTP 404): a completeness gap in R2, or a
+    /// transient miss near the tip / right after a reorg, before the uploader has PUT the object.
+    #[error("R2 witness MISSING for block {number} (key {key}): object not found (404)")]
+    Missing { number: u64, key: String },
```
vincent-k2026, line 67: **Low — `Missing` loops forever (even with `--end-block`)**

Near tip, infinite retry + 2s throttle is correct (uploader race). For historical slice runs, a permanently missing object means the process never finishes and never fails.

Worth documenting clearly for operators, and optionally fail-fast when `sync_target` is set after N consecutive `Missing` for the same `(number, hash)` so batch jobs surface a completeness gap instead of hanging.

bin/stateless-validator/src/r2_witness.rs
```diff
@@ -0,0 +1,583 @@
...
+        if body.len() > MAX_ERROR_BODY_BYTES {
+            let mut end = MAX_ERROR_BODY_BYTES;
+            while !body.is_char_boundary(end) {
+                end -= 1;
+            }
+            body.truncate(end);
+        }
+        // A 404 usually means the object is absent (`NoSuchKey`) — but S3 also 404s a missing
+        // *bucket*, which is operator misconfiguration, not a data gap; keep those apart.
+        if code == 404 && !body.contains("NoSuchBucket") {
```
vincent-k2026: **Low — `NoSuchBucket` via `body.contains`**

Works for real S3 XML error bodies. Slightly brittle if a proxy returns HTML (or truncation ever eats the `<Code>`). Prefer matching a small set of known S3 error codes (e.g. parse `<Code>…</Code>` / check for `NoSuchKey` positively for `Missing`) if you touch this again — not blocking.

vincent-k2026 (approved):

PR #153, opened 2026-07-11
feat: coverage-replayer — minimal mainnet block set maximizing mega-evm branch coverage

flyq:
## Summary
Adds `bin/coverage-replayer`, an offline tool that replays mainnet blocks under LLVM branch instrumentation and derives the minimal block set covering every branch counter ever observed in mega-evm — the source of truth for keeping `test_data/mainnet` maximally representative. Also adds the zero-validation "light" witness decode to stateless-core/common and switches debug-trace-server onto it. Carries PR #152 (stateless-r2) via a merge; those files are reviewed there.

## coverage-replayer
`backfill` fetches blocks + witnesses (RPC or straight from R2 with `--witness-source r2`), resident worker subprocesses replay each block with per-block counter isolation (reset → `replay_block` → capture; process isolation keeps the dispatcher out of the bitmaps), and a judge dedups per-block coverage bitmaps into patterns in a redb store, keeping one ~106KB sparse profdata per non-dominated pattern. Never-skip semantics: fetch retries forever, slow blocks are warned about but never killed, replay errors / header-sanity divergence fail-stop the run and are retried on re-run. `set-cover` computes the greedy minimal cover (antichain pruning, incumbent-biased tie-breaks, redundancy elimination); `report` renders llvm-cov for the selected set; `merge` combines per-machine shard stores (dense remap via content-addressed counter ids); `inspect` prints store statistics. Stores are namespaced by a build fingerprint (mega-evm rev + toolchain/host) and the symbol filter.

## Light witness decode
`LightWitnessFromSalt` (stateless-core) decodes kvs+levels directly from full `SaltWitness` bytes via a layout-mirroring deserializer — no curve points constructed: ~1.4ms vs ~110ms wall (~700x CPU) on a 6.3MiB mainnet witness. Locked by byte-exact tests against every mainnet fixture and the committed `test_data/mainnet/bench/6906405.zst`. `get_witness_light` / `decode_witness_payload_light` in stateless-common.

## debug-trace-server
Both RPC witness paths switch to the light decode (the server never verifies proofs): uncached trace requests lose ~100ms on large witnesses; chain-sync prefetch stops burning ~1 core·s/block. Note: the `witness_generator` size metric now records a documented lower bound (excludes parent commitments), shifting the histogram down ~40-60%. Unused `salt` dep dropped.

## Verification
267 workspace tests green (incl. new set-cover/judge/merge/R2/light-decode units and a fixture integration test replaying all `test_data/mainnet` blocks with gas/receipts/bloom equality). Validated on mainnet: 241k blocks replayed with zero sanity failures; branches coverage 47.90% for the selected 45-block set vs 32.63% from tip-only traffic.


flyq:
Superseded by #222. This branch predated several large refactors on `main` (notably the #152 squash-merge and the validator restructure), so its diff had become dominated by content `main` already carries — 49 files where the tool itself is 22. #222 is `main` plus the tool only, ported to mega-evm v1.7.0, with the full-history scan results in its description.

PR #155, opened 2026-07-17
feat(stateless-core): validate_block_deriving_updates — changeset-deriving validation entry point

flyq:
## Summary
Adds `validate_block_deriving_updates`, a `validate_block` variant that returns the replay-derived `salt::StateUpdates` instead of computing the post state root. It targets embedders that already hold an independently verified per-block changeset for the same block — a MegaETH full node, whose state sync hash-verifies the sequencer's `SaltDeltas` against the signed header (`salt_deltas_hash`) and recomputes/checks the post state root at commit time. Comparing the replay-derived net update map against that stored changeset replaces the validator-side SALT trie/commitment recompute (~0.65s/block measured), which duplicates work state sync has already done.

## Design
- `validate_block_deriving_updates(chain_spec, block, salt_witness, mpt_witness, contracts, options, writer)` → `(StateUpdates, ValidationStats)`. Both witness proofs are **always verified** (the SALT witness's IPA proof before replay, the MPT witness's Merkle proof on the replay outputs); no post state root is computed — the caller owns the changeset comparison.
- `ValidationOptions` (`#[non_exhaustive]`, deliberately no `Default`) is built via `ValidationOptions::anchored(state_root, withdrawals_root)` — the standard form, binding both witnesses' pre-roots to the parent's post-root pair before any other work (`PreStateRootMismatch` / `PreWithdrawalsRootMismatch` on divergence) — or the explicit `ValidationOptions::unanchored()` opt-out for callers that genuinely lack a parent header. The `ParentAnchor` pair mirrors the `(state root, withdrawals root)` pair the standalone pipeline's continuity check enforces and is atomic by construction: real callers take both roots from the same parent header and never hold one without the other. The anchor is the only check binding the MPT witness's *pre*-state to the chain — `MptWitness::verify` proves the witness against its own claimed root and binds only the *post* root to the header.
- The replayed header claims (withdrawals root via MPT witness, receipts root, logs bloom, gas used) are checked exactly as in `validate_block`, before the state-update derivation.
- The shared stages (`verify_and_replay`, `withdrawal_storage`, `derive_state_updates`, `verify_replay_outputs`, `timed`) are extracted and reused by `validate_block`, whose behavior is unchanged (`crates/stateless-core/src/executor.rs`).
- On success `ValidationStats::salt_update_time` holds only the state-update derivation time (no trie math happens here) — documented on the field.

## Why strict changeset equality is sound
`salt::StateUpdates` is a canonical net `{key ↦ (old, new)}` map uniquely determined by the (pre, post) state endpoints — order-independent, no-op-free. The embedder's stored row is RLP-hash-committed in the signed header and its application to the pre-state is root-verified by state sync at commit; the replay derivation here produces the net map of the same transition, so field-for-field equality on puts/deletes/updates is exactly the ingest-time-enforced equation (`codes` is not derivable from state updates and stays covered by the ingest-time header-hash and bytecode checks).

## Review changes
- Renamed from `validate_block_updates` (review: the updates are the function's output, not the object of validation; the new name matches the internal `derive_state_updates` stage and keeps the `validate_block_` prefix signalling unconditional proof verification).
- Anchoring hardened across review rounds: single state-root anchor → atomic `ParentAnchor` pair (the withdrawals pre-root was previously unbound) → silent `Default` removed in favor of explicit `anchored`/`unanchored` constructors, so skipping the parent binding must be spelled out at the call site.

## History note
Earlier revisions of this branch also introduced `validate_block_updates_light` (changeset validation over the zero-validation light witness, no IPA and no anchor) plus a `verify_witness` opt-out. Both were removed in b229531 on security-assumption grounds — proof verification is now unconditional. The light witness decode itself (#154, merged) is untouched and still serves debug-trace-server.

## Testing
`cargo nextest run --workspace` — 264/264 pass; the CI no-std job command `cargo test -p stateless-core --no-default-features --lib` — 54/54 locally. Key tests: the mainnet fixture sweep runs anchored to each parent (the embedder's real call shape) and cross-checks that the returned updates reproduce the header state root through the SALT trie update, locking equivalence with `validate_block`; the paired-anchor test asserts the witness-root/parent-root invariant across all paired fixtures and rejects each mismatched half with exact error fields, using a tampered witness to prove the anchor runs before proof verification; `tampered_witness_fails_proof_verification` locks the unconditional IPA check through both entry points.

### DIFF
--- a/crates/stateless-core/src/executor.rs
+++ b/crates/stateless-core/src/executor.rs
@@ -10,6 +10,8 @@
 //!   transaction replay, and state root comparison
 //! - [`validate_block_updates`]: Variant returning the replay-derived SALT state updates for
 //!   embedders that compare against an independently verified per-block changeset
+//! - [`validate_block_updates_light`]: The same over a zero-validation [`LightWitness`] — no curve
+//!   point is ever constructed on the whole decode + replay path
 //! - [`create_evm_env`]: Creates EVM execution environment from block header and chain
 //!   specification
 //! - [`replay_block`]: Replays block transactions to compute state changes
@@ -59,7 +61,10 @@ use revm::{
     primitives::{B256, KECCAK_EMPTY, U256},
     state::Bytecode,
 };
-use salt::{EphemeralSaltState, SaltValue, SaltWitness, StateRoot, StateUpdates, Witness};
+use salt::{
+    EphemeralSaltState, SaltValue, SaltWitness, StateRoot, StateUpdates, Witness,
+    traits::StateReader,
+};
 use thiserror::Error;
 use tracing::debug;
 
@@ -67,6 +72,7 @@ use crate::{
     chain_spec::{BLOB_GASPRICE_UPDATE_FRACTION, ChainSpec},
     data_types::{Account, PlainKey, PlainValue},
     evm_database::{WitnessDatabase, WitnessDatabaseError, WitnessExternalEnv},
+    light_witness::{LightWitness, LightWitnessError, LightWitnessExecutor},
     withdrawals::{self, ADDRESS_L2_TO_L1_MESSAGE_PASSER, MptWitness},
 };
 
@@ -90,6 +96,9 @@ pub enum ValidationError {
 
     #[error("Failed to update salt state: {0}")]
     StateUpdateFailed(#[source] salt::SaltError),
+
+    #[error("Failed to update salt state over a light witness: {0}")]
+    LightStateUpdateFailed(#[source] LightWitnessError),
 
     #[error("Failed to update salt trie: {0}")]
     TrieUpdateFailed(#[source] salt::SaltError),
@@ -467,10 +476,15 @@ fn withdrawal_storage(accounts: &HashMap<Address, BundleAccount>) -> B256Map<U25
 ///
 /// The result is the net `{key ↦ (old, new)}` map between the block's pre- and post-states —
 /// the same map the SALT trie update consumes and the sequencer's `SaltDeltas` are derived from.
-fn derive_state_updates(
-    witness: &Witness,
+///
+/// Generic over the witness store: `salt::Witness` on the proof-carrying paths
+/// ([`validate_block`] / [`validate_block_updates`]) and [`LightWitnessExecutor`] on the
+/// zero-validation path ([`validate_block_updates_light`]); callers map the store error into
+/// their [`ValidationError`] variant.
+fn derive_state_updates<W: StateReader>(
+    witness: &W,
     accounts: HashMap<Address, BundleAccount>,
-) -> Result<StateUpdates, ValidationError> {
+) -> Result<StateUpdates, W::Error> {
     // Flatten Revm's BundleAccount format into plain key-value pairs
     let mut kv_updates: BTreeMap<Vec<u8>, Option<Vec<u8>>> = BTreeMap::new();
     for (address, bundle_account) in accounts {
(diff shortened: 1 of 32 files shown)


crates/stateless-core/src/executor.rs
```diff
@@ -626,21 +775,184 @@ pub fn validate_block<B: BlockInput>(
...
+///   bytecode integrity must be enforced at ingest.
+/// - [`ValidationOptions::parent_anchor`], when provided, is checked against both witnesses' own
+///   pre-roots before any other work, anchoring them to the canonical parent block — see its field
+///   docs for what each half binds and which error it raises.
+///
+/// Both witness proofs (the SALT witness's IPA proof before replay, the MPT witness's Merkle
+/// proof on the replay outputs) are always verified, exactly as in [`validate_block`].
+/// On success, [`ValidationStats::salt_update_time`] holds the state-update derivation time
+/// (there is no trie update here).
+pub fn validate_block_updates<B: BlockInput>(
```
Troublor: The name `validate_block_updates` reads as "validate the block-updates," but the updates are the *output* — the function validates the block and derives the updates, while the updates-vs-changeset comparison belongs to the caller (as the doc itself notes). The doc having to clarify "the caller owns that comparison" suggests the name isn't carrying that meaning on its own.

Consider `validate_block_deriving_updates`: it keeps the `validate_block_` prefix (signaling this is a full-trust entry point with unconditional proof verification) and matches the internal `derive_state_updates` stage, while making the updates unambiguously the return value. `validate_block_into_updates` would also work.

(I'd avoid `replay_*` names — they'd read as siblings of the proof-skipping `replay_block`, which is a worse confusion.)

Since this is the PR's main public export, a rename is cheap now but expensive after merge — worth settling before approval.

Troublor (changes requested):
Solid PR overall: the stage extraction is pure code motion (validate_block behavior and error precedence unchanged), the ParentAnchor design is sound, and the test suite is strong — the trie-based equivalence cross-check in particular locks the two entry points together.

Requesting changes on one point: the name of the main new export (see inline comment). Since it's the PR's public API surface, it's worth settling before merge.

One more note: the PR description still describes `ValidationOptions { expected_pre_state_root }`, but the code now anchors a `(state_root, withdrawals_root)` pair via `parent_anchor`. Since the description becomes the squash commit message, please update it before merging.

crates/stateless-core/src/executor.rs
flyq: fixed

Troublor (approved):

vincent-k2026 (approved):
Approving. Independent pass over the stage extraction, the anchor design, and the test suite — no blocking issues found.

Verified:
- `validate_block` extraction is pure code motion: error precedence (`BlockIncomplete → EnvOracle → WitnessVerification → replay → TrieUpdateFailed → withdrawals/receipts/logs/gas → StateRootMismatch`) and the per-stage timing split are preserved; `salt_update_time` now excludes the (µs) withdrawal-storage extraction, as documented.
- `parent_anchor` is the sole binding of either witness's *pre*-state to the canonical parent — `MptWitness::verify` only self-checks against its own `storage_root` and binds the *post* root to the header, so the withdrawals anchor is load-bearing, not defense-in-depth. Sound.
- Tests discriminate: the tampered-witness + bogus-anchor case proves anchor-before-verify ordering with field-exact errors, and the equivalence test reproduces the header state root through the trie to lock both entry points together.

One non-blocking, cross-repo follow-up (not this PR): the returned `StateUpdates` cannot cover newly deployed bytecode (documented on the fn), so the mega-reth consumer must keep enforcing bytecode integrity at ingest rather than treating the changeset comparison as complete validation.

(I did not re-run the suite locally — salt is a git dep / cold build; relied on green CI + a static read of `executor.rs` and `withdrawals.rs`.)

PR #156, opened 2026-07-22
feat(metrics): per-endpoint RPC attempt metrics + timeout reasons

vincent-k2026:
## Summary

Witness fetches hit the primary (witness-generator) and fall back to Cloudflare/R2 under one shared 3s deadline, but the metrics collapsed every provider — and every failure reason — into a single `method`-keyed series, and the retry loop's failure logs named neither the endpoint nor the block. There was no way to answer "did this 3s go to the witness-generator or the failover?", "how many were timeouts vs errors?", or "which block stalled on which endpoint?".

This PR attributes upstream RPC **latency, failure reason, timeout counts, and per-attempt logs per endpoint**, for any number of fallback endpoints. The mechanism lives in the shared `round_robin_with_backoff` retry loop, so both binaries benefit; the debug-trace-server (the witness path the on-call was investigating) gets the full per-endpoint Prometheus treatment.

## What changed

**`stateless-common` (shared retry loop + trait)**
- `RpcMetrics::on_rpc_complete(method, success, duration)` → `on_rpc_attempt(method, provider, outcome, duration)`, where `outcome` is a new `RpcAttemptOutcome { Success, Error, Timeout }` — splits a provider that *returned an error* from one that *stalled and had to be timed out*.
- New `on_rpc_deadline_exceeded(method, elapsed)` (default no-op) — the operator-facing "the whole logical call blew its deadline" signal, recorded exactly once per call.
- `round_robin_with_backoff` classifies each attempt, tags it with a bounded, credential-stripped endpoint label (`endpoint_label`, format `{idx}:{host}`, strips `user:pass@`/path/query), and threads that label into the per-attempt WARN/DEBUG logs.
- **`timeout` counts genuine provider stalls only** — a timeout that fired merely because the caller's deadline clamped the attempt window (deadline pressure, not an upstream stall) is attributed to `deadline_exceeded`, not blamed on the provider.
- Successful witness fetches log the serving endpoint and split `request_ms` (RPC round trip) from `decode_ms` (CPU decode).
- A `warn`-level span carries `block_number`/`block_hash` around the witness retry, so the generic loop's per-attempt failure logs inherit the block identifier (they have no block context of their own).
- `set_validated_blocks` (single attempt, outside the loop) classifies its own outcome the same way.

**`debug-trace-server`**
- Replaced the method-only `UpstreamMetrics` (a `success` bool) with `(method, provider, outcome)`-keyed attempt series plus a per-method deadline-exceeded counter.

**`stateless-validator`**
- Adapted `ValidatorMetrics` to the new trait; added `stateless_validator_rpc_deadline_exceeded_total{method}`.

## Metrics

debug-trace-server:
- `debug_trace_upstream_requests_total{method, provider, outcome}` — one per provider attempt that ran its course (`outcome` ∈ `success|error|timeout`).
- `debug_trace_upstream_duration_seconds{method, provider, outcome}` — per-attempt latency, per endpoint.
- `debug_trace_upstream_deadline_exceeded_total{method}` — **new**: logical calls that exhausted their deadline budget (also absorbs deadline-pressure timeouts).

stateless-validator:
- `stateless_validator_rpc_deadline_exceeded_total{method}` — **new**.

`provider` is `{idx}:{host}` — the host is human-readable, and the leading index keeps two endpoints that share an authority distinct while encoding primary (0) vs failover (1+). So `sum by (provider)` gives the per-endpoint split for any number of fallbacks.

## Logs

Per-attempt failure logs (stall / error) carry `method`, `provider=<idx:host>`, `provider_idx`, and — via the span — `block_number`. Successful witness fetches trace `provider`, `request_ms`, `decode_ms`. So a stall reads directly: which block, which endpoint, which method, how long. (A stall now logs once with its reason instead of also emitting a redundant "trying next".)

## Operational note (dashboards)

Same class of change as #154 — dashboards/alerts need a small rebase when this deploys:
- `debug_trace_upstream_errors_total` is **removed**; error count is now `debug_trace_upstream_requests_total{outcome=~"error|timeout"}`, timeouts separable as `{outcome="timeout"}` (genuine stalls; deadline pressure lives in `debug_trace_upstream_deadline_exceeded_total`).
- `debug_trace_upstream_requests_total` / `debug_trace_upstream_duration_seconds` gained `provider` + `outcome` labels — aggregating queries keep working; add `by (...)` where needed.
- Per-attempt upstream series carry a dynamic `provider` label, so they appear on first attempt; the deadline-exceeded counter is pre-registered per method.

## Testing

- `cargo test --workspace`: passes (network integration tests ignored as usual); `fmt` / `clippy --workspace --all-targets --all-features` clean.
- New tests: `endpoint_label` host extraction + credential/userinfo stripping + same-host `{idx}` disambiguation; `RpcAttemptOutcome`; a capturing `RpcMetrics` mock asserting per-endpoint attempt outcomes on a primary-fails/backup-succeeds failover and exactly one `on_rpc_deadline_exceeded` per deadline-exhausted call; and a captured-log test asserting the stall WARN carries `block_number` at the default `warn` filter (deadline-independent so it can't flake on CI timing).

## Notes

- `mega-reth` consumes `stateless-common` but does not implement `RpcMetrics`, so the trait change is contained to this repo; it needs no change when it next bumps the `stateless-common` tag.
- `provider_labels` are parallel-by-index to `providers`; a `debug_assert_eq!` at the retry-loop entry makes that invariant explicit.

### DIFF
--- a/bin/debug-trace-server/src/data_provider.rs
+++ b/bin/debug-trace-server/src/data_provider.rs
@@ -767,7 +767,7 @@ async fn resolve_contracts_inner(
         "Cache miss — fetching contracts from RPC"
     );
 
-    // Per-attempt `eth_getCodeByHash` metrics land on `UpstreamMetrics` via the
+    // Per-attempt `eth_getCodeByHash` metrics land on the upstream attempt metrics via the
     // `TraceRpcMetrics` adapter inside `round_robin_with_backoff`.
     let fetched = rpc_client.get_codes_with_deadline(&missing, true, Some(deadline)).await?;
 
--- a/bin/debug-trace-server/src/metrics.rs
+++ b/bin/debug-trace-server/src/metrics.rs
@@ -8,7 +8,7 @@
 use std::net::SocketAddr;
 
 use eyre::Result;
-use metrics::{Counter, Gauge, Histogram};
+use metrics::{Counter, Gauge, Histogram, counter, histogram};
 use metrics_derive::Metrics;
 use metrics_exporter_prometheus::{Matcher, PrometheusBuilder};
 pub use stateless_common::{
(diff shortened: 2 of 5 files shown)


Troublor (approved):

flyq (approved):
LGTM

PR #157, opened 2026-07-22
docs: add stateless-validator review checks to REVIEW.md

vincent-k2026:
Supplements the baseline rubric with stateless-validator-specific checks it doesn't cover:

- Fail closed on the terminal publish — a permanently failed final step surfaces a non-zero exit / `Err`, not silent success.
- Validate config at the boundary — reject/clamp a `0` timeout and unbounded alloc hints before they reach the retry loop.
- Classify remote errors precisely — match known error codes (not `body.contains`), with an inter-cycle retry throttle.

Additive only — no existing rules changed.

### DIFF
--- a/REVIEW.md
+++ b/REVIEW.md
@@ -17,6 +17,9 @@ These per-repo rules take precedence wherever they conflict with that baseline.
 
 - Changes to consensus-critical code — EVM execution, witness verification, state root computation, withdrawals — require extra scrutiny.
 - Watch for read-check-write without atomicity (lock, CAS, or database constraint), CPU-intensive work that should use `spawn_blocking` inside async contexts, and deadlocks from nested lock acquisition.
+- **Fail closed on the terminal publish.** The final step (e.g. `mega_setValidatedBlocks`) must not look like success when it permanently failed — surface a non-zero exit / `Err` so the orchestrator detects the miss and retries, instead of advancing while upstream never saw the tip.
+- **Validate config at the boundary.** Reject or clamp nonsensical config at load: a `*_timeout_ms` of `0` makes `tokio::time::timeout` fail every attempt immediately; an unbounded size/alloc hint invites upfront over-allocation. Require `> 0` (or a clamp) before the value reaches the retry loop.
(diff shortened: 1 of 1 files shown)


REVIEW.md
```diff
@@ -17,6 +17,9 @@ These per-repo rules take precedence wherever they conflict with that baseline.
 
 - Changes to consensus-critical code — EVM execution, witness verification, state root computation, withdrawals — require extra scrutiny.
 - Watch for read-check-write without atomicity (lock, CAS, or database constraint), CPU-intensive work that should use `spawn_blocking` inside async contexts, and deadlocks from nested lock acquisition.
+- **Fail closed on the terminal publish.** The final step (e.g. `mega_setValidatedBlocks`) must not look like success when it permanently failed — surface a non-zero exit / `Err` so the orchestrator detects the miss and retries, instead of advancing while upstream never saw the tip.
```
flyq: - **Tighten "permanently failed" in rule 1** (`REVIEW.md:20`). The codebase deliberately tolerates a failed final report on a *chain-following* run — the next start re-reports anchor→tip and the miss heals (`workers.rs:170-178`). Only the slice run fails closed. As written, the rule relies entirely on the word "permanently" to encode that distinction, and the automated reviewer reads this line without the `workers.rs` context — it could flag the deliberately-tolerant chain-following path as a violation. Suggest: "…must not look like success when the run has no later opportunity to retry (e.g. an `--end-block` slice run)".

flyq (approved):

REVIEW.md
vincent-k2026: Addressed in 7fae927. Reworded to key the fail-closed requirement on "the run has no later opportunity to retry (e.g. an `--end-block` slice run)" instead of leaning on the word "permanently", and called out the chain-following run (re-reports anchor→tip on its next start, so the miss self-heals) as the deliberate exception — matching `workers.rs:170-178`. Thanks for the catch!

flyq (approved):
LGTM

Troublor (approved):

PR #161, opened 2026-07-27
refactor(debug-trace-server): hash-keyed response cache + in-memory canonical-hash memo

flyq:
## Summary
Restructures the response cache along the direction set by the review (thanks @vincent-k2026 — both rounds, taken wholesale). The cache key becomes `(resource, block_hash, typed_variant)`: an entry is an immutable fact about one exact block, so the canonicality scaffolding (`is_canonical`, re-key refusal, stale purging, depth-trust, both race windows) is deleted rather than patched. By-number handlers resolve number → canonical hash *before* the cache lookup: the bounded `CANONICAL_CHAIN` window, then a **bounded in-memory canonical-hash memo**, then upstream `eth_getHeaderByNumber` (hash-verified) — round 2's memo suggestion adopted after the production replay priced the alternatives. `CANONICAL_CHAIN` keeps its original bounded, contiguous-by-construction semantics — chain-sync and reorg-bisection invariants are untouched — and the PR adds **no new on-disk surface**: no tables, no migration, downgrade-safe. (Earlier revisions used a bulk backfill, a permanent chain with a history floor, and a dedicated `HASH_ARCHIVE` table; all superseded — the `data_provider.rs:331` thread has the data that decided it.)

## Review items → what happened
- **R1-B1 — unbounded key space / 64-bit collision**: `tracerConfig` is never hashed as raw JSON. `RequestShape::classify` parses it once into alloy's typed configs and the key carries Hash-able mirrors (`CallConfigKey`/`PreStateConfigKey`/`FlatCallConfigKey`), so the per-tracer key space collapses to a closed ~60-value set: no flooding, nothing to birthday-search, junk/reordered/null configs merge onto one entry. `4byteTracer`/`noopTracer` carry no config at all.
- **R1-B2 — slot written outside the index lock**: structurally gone — the key contains the hash, so there is no `number → hash → slot` indirection left to race. The secondary index is only `hash → keys` for reorg/eviction cleanup, registered before the cache write (immediate weight-eviction self-heals via `on_evict`; an `invalidate_blocks` landing between the two steps leaves a harmless unindexed entry, documented).
- **R2-N1 — post-reset write-back could permanently pin a non-canonical block**: structurally impossible now — a reorg-able height is never memoized (`block_num < observed_tip − 64`, where the tip reference is a monotonic max of verified on-chain heights and never exceeds the real tip), and a memo entry cannot outlive the process.
- **R2-N2 — redb writer lock inline on the async request path**: no DB write exists on the request path anymore.
- **R2 main ask — in-memory memo over persistence**: adopted. The 3.6-day replay put below-retention by-number traffic at ≈79% in the storm window vs ≈4–5% steady state, with a ~50–130k-unique-height working set — comfortably inside an LRU; the archive's increments (restart persistence, pruning feed) priced out small against permanent-poisoning blast radius and monotonic disk growth into the size-prune ratchet. `debug_trace_canonical_hash_resolution_total{source="memo",outcome="miss"}` now measures exactly the volume that would justify revisiting persistence as its own PR.
- **`is_canonical` fail-open, depth-trust, unmeasured tri-state, fixed 2 s deadline**: the function no longer exists. By-number handlers resolve the hash on the request's own shared deadline from `fetch_deadline()` — which also fixes the pre-existing double-budget (tag resolution and fetch each minting 13 s). Resolution is counted by `debug_trace_canonical_hash_resolution_total{source="db"|"memo"|"upstream", outcome}`; DB read errors are logged and fall through instead of being swallowed.
- **`shape_label` / `cacheable_from_geth_options` drift**: replaced by the single `RequestShape::classify` with `label()`/`cache_variant()` derived from it; the test pins the classification, not a case list.
- **`unwrap_or_default()` on malformed tracerConfig**: a type-malformed config on call/prestate/flatCall is rejected with `-32602 invalid params` at the handler layer (`classify_and_gate`, also on `debug_traceTransaction`) before any fetch or execution.
- **`estimated_items = 0` as off-switch**: `--response-cache-disabled` (+ env) is the explicit switch; `0` is rejected at parse time, CLI and env.
- **Test hygiene**: positive coverage via in-process jsonrpsee mocks serving self-consistent (`hash_slow`-derived) headers, fast-fail retry configs, `let _listener = listener;`; the 2 s/4 s canonicality tests are gone with the code they tested.
- **Nits** (`get_or_compute` dead code, `TracerType::parse`, `insert_cache` twins, `variant?.clone()`, `&key != keep`): all deleted along with their host code paths.
- **Deployed-mode question**: fra and ore both run local-cache mode, so steady-state by-number resolution is a local DB read; stateless mode pays one upstream header fetch per cold height and then hits the memo.

## Canonical-hash memo
Resolution order in `resolve_canonical_hash`: the bounded sync window, then the memo, then upstream with `verify_hash = true` (the client recomputes `hash_slow`, so the binding provably matches the header content). An upstream answer is memoized only when depth-final — more than `CANONICAL_MEMO_MIN_DEPTH` (64) below the observed tip — so a memoized binding can no longer reorg; shallow and above-tip heights resolve upstream every time. The tip reference is a monotonic max of observed on-chain heights (latest-tag resolutions, verified headers, the local window tip): it can only under-estimate the real tip, so the gate errs conservative, and it works in stateless mode too. Capacity is `--canonical-hash-memo-capacity` (default 8M entries ≈ 0.6 GB if ever full; the memo fills lazily, so the cap costs nothing until a deep historical scan uses it; `0` rejected at parse time). First touch of a cold height pays one upstream RTT; with production's 93.7 % repeat-after-success rate, everything after that is a local read for the process lifetime. A restart clears the memo — rewarm is ~200 RTTs/h at steady-state traffic, and during a storm it warms at the storm's own pace.

## Behavior changes
- Malformed `tracerConfig` on call/prestate/flatCall → `-32602` (previously traced silently with defaults, HTTP 200), including `debug_traceTransaction`.
- `--response-cache-estimated-items 0` refuses to start (CLI and env). **Deployment note:** envs still exporting `DEBUG_TRACE_SERVER_RESPONSE_CACHE_ESTIMATED_ITEMS=0` fail loudly after upgrade until switched to `DEBUG_TRACE_SERVER_RESPONSE_CACHE_DISABLED=true` (or a real item count) — intended forcing function.
- New `--canonical-hash-memo-capacity` (env `DEBUG_TRACE_SERVER_CANONICAL_HASH_MEMO_CAPACITY`), default 8,000,000, must be ≥ 1.
- Size-based pruning respects `--size-prune-min-retain` (default 256), so a DB file permanently over `--db-max-size` can no longer consume the entire body retention; because redb files never shrink, crossing the limit ratchets body retention to that floor until `--db-max-size` is raised or the DB is rebuilt (documented).
- New `db_body_earliest_block` gauge (body-retention edge); `db_earliest_block` keeps meaning the sync window's start.
- No DB schema change in either direction: no new tables, zero migration, downgrade-safe.

## Testing
`cargo test --workspace` green: 85 debug-trace-server unit tests (incl. memo end-to-end: resolve upstream → upstream killed → depth-final height served from the memo while a shallow height is never memoized and fails; memo-capacity CLI/env/zero-rejection coverage; prune/reset behavior), 17 stateless-db tests, validator suite untouched and green. `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features` (0 warnings), `cargo sort --check` all pass.

## Out of scope (deliberate)
PR #162 rebase (deprioritized per request-share data), `redb::Builder::set_cache_size` tuning, offline `compact()`, response-level request coalescing, single-flighting of cold canonical-hash resolutions, the pre-existing mux config-parse inconsistency, the pre-existing `-h`/`--version` CLI arg bug, and a persistent number → hash index (revisit as its own PR if `source="memo"` miss telemetry ever shows the volume to justify it).


bin/debug-trace-server/src/response_cache.rs
```diff
@@ -286,6 +326,15 @@ impl SecondaryIndices {
...
 }
 
+/// Outcome of [`SecondaryIndices::insert`].
+enum IndexInsert {
+    /// Mapping and key recorded; on a canonical re-key, carries the dead block's keys.
+    Inserted(Option<HashSet<ResponseCacheKey>>),
+    /// The number is mapped to a different hash and re-keying was not allowed; nothing was
+    /// recorded and the caller must not cache the response.
+    Rejected,
+}
```
Troublor: `Inserted(Option<HashSet<..>>)` folds two different outcomes into one variant: "recorded, nothing displaced" and "recorded after re-keying, here are the dead block's keys". The distinction is the whole reason this type exists, but it lives in the payload rather than the variant name, so every call site has to re-derive it — and `evict_stale_keys` takes an `Option` whose `None` case is a no-op it must guard.

Naming the third outcome makes each arm self-describing:

```suggestion
/// Outcome of [`SecondaryIndices::insert`].
enum IndexInsert {
    /// Mapping and key recorded; the number had no conflicting mapping.
    Inserted,
    /// Mapping and key recorded after re-keying the number to a new hash; carries the
    /// dead block's keys for eviction.
    Rekeyed(HashSet<ResponseCacheKey>),
    /// The number is mapped to a different hash and re-keying was not allowed; nothing was
    /// recorded and the caller must not cache the response.
    Rejected,
}
```

Call sites then read as the three real cases, and `evict_stale_keys` can take a plain `HashSet` (dropping its `let Some(stale) = stale else { return }` guard):

```rust
match self.inner.indices.insert(block_hash, block_number, key.clone(), true) {
    IndexInsert::Inserted => {}
    IndexInsert::Rekeyed(stale) => self.evict_stale_keys(stale, &key),
    IndexInsert::Rejected => unreachable!("canonical inserts may always re-key"),
}
```

Non-blocking, but worth it while the type is new.

bin/debug-trace-server/src/response_cache.rs
```diff
@@ -131,28 +111,81 @@ pub enum ResponseVariant {
...
-        } else {
             let mut hasher = DefaultHasher::new();
-            // Hash the debug representation of the config
             format!("{:?}", opts.tracer_config).hash(&mut hasher);
-            // Also hash relevant default options
-            format!("{:?}", opts.config).hash(&mut hasher);
-            Some(hasher.finish())
-        };
+            hasher.finish()
+        });
```
vincent-k2026: **Blocking — the whitelist bounds *which shapes* may be cached, but not *how many keys* they can mint.**

`tracer_config` is fully caller-controlled JSON, and it enters the cache key as a 64-bit `DefaultHasher` digest of the **raw JSON's `Debug` string**. None of `CallConfig` (alloy `.../geth/call.rs:66`), `FlatCallConfig` (`:99`) or `PreStateConfig` (`geth/pre_state.rs:206`) carries `deny_unknown_fields` — that attribute only sits on the *output* type `DiffMode` — and `into_*_config()` goes through `serde_json::from_value`, which silently ignores unknown fields. Three consequences:

1. **Unbounded key space (the one I'd actually worry about).** `{"tracer":"callTracer","tracerConfig":{"pad":i}}` for i = 1, 2, 3… mints unlimited distinct keys whose responses are **byte-identical** to bare `callTracer`. That evicts the whole 1 GB weight-based LRU and drops everyone back to recomputation, at the same marginal cost to the caller as any normal trace request. For a PR whose stated purpose is hardening the cache before its first production enable, this is the door the whitelist should have closed.
2. **Collision.** Free padding makes a 64-bit birthday search (~2^32, offline) practical, e.g. landing `onlyTopCall:true` and `onlyTopCall:false` on the same entry — a **silently wrong trace**. Note the inverse: if those three types were `deny_unknown_fields`, `CallConfig` would only have 9 semantic values and there would be nothing to search. It is "ignore unknown fields" that turns the collision from infeasible into feasible.
3. **Hit-rate loss.** Semantically identical configs that differ in key order or carry irrelevant fields each occupy their own entry. (This workspace does enable `serde_json`'s `preserve_order` — visible in `cargo tree -p debug-trace-server -e features -i serde_json` — so `{:?}` is key-order sensitive.)

**Suggested fix — one change covers all three:** don't hash the raw JSON. Parse it into the tracer's typed config first and put the **parsed value** into `ResponseVariant`. `CallConfig` / `PreStateConfig` / `FlatCallConfig` are `Copy + PartialEq + Eq` structs of 2–3 `Option<bool>`s, so the whole variant space collapses from unbounded to ~60 combinations: no collisions, no key flooding, and equivalent configs merge automatically. `4byteTracer` / `noopTracer` never read `tracer_config` at all, so their variant should drop the config outright rather than hash an irrelevant blob.

Related (pre-existing, out of scope, but the same code path): `tracing_executor.rs:467 / 478 / 489` use `into_call_config().unwrap_or_default()`, so a type-level parse error (e.g. `{"diffMode":"true"}`) silently degrades to the default config — a `diffMode` request gets non-diff output with HTTP 200, where geth would error. If you do the parse-then-key change, making that a client error costs almost nothing extra.

bin/debug-trace-server/src/response_cache.rs
```diff
@@ -435,10 +552,53 @@ impl ResponseCache {
         let cached = CachedResponse::new(response);
 
         self.inner.cache.insert(key.clone(), cached);
-        self.inner.indices.insert(block_hash, block_number, key);
+        match self.inner.indices.insert(block_hash, block_number, key.clone(), true) {
```
vincent-k2026: **Blocking — cache slot is written before the index is re-keyed, so a by-hash read can be served another block's response.**

`get_or_compute` has the same ordering (`:638` then `:640`). `get_by_hash` (`:519-529`) has no verification at all — it trusts the `hash → number → slot` chain, but the slot is hash-agnostic:

```
initial: cache[(N,v)] = respA,  index: A→N, N→A
thread A: insert(N, hashB, v, respB)  → :554 cache[(N,v)] = respB
thread B: get_by_hash(A)              → :526 resolves N (index not yet re-keyed — this step is correct)
                                      → :529 reads respB              <-- wrong answer
thread A:                             → :555 re-key, drops A→N
```

A caller who explicitly asked for hash A gets hash B's trace. The reader being preempted between `:526` and `:529` widens the window beyond the two writer statements.

To be precise about what this relies on: **not** "one hash mapping to several numbers" (that indeed cannot happen), but **one number mapping to several hashes over time** — which is exactly the premise `IndexInsert::Rekeyed`, `insert_by_hash`'s `Rejected` and `is_canonical` exist for. The root cause is that `SecondaryIndices` maintains the bijection under one write lock, but **the cache slot write lives outside that lock** (as this PR already documents — sharing a lock with quick_cache's eviction callback would deadlock). So "number is a proxy for hash" holds inside the index lock and not on the slot; `get_by_hash` can never confirm the slot was written under the hash it was asked about, because the key carries no hash.

Codex's `:533` comment was the mirror image on `insert_by_hash`, and the after-write re-check narrowed that one; this side (`insert` / `get_or_compute` ordering, plus the **read** path) has no equivalent guard.

**Suggested fix:** the real fix is putting the hash in the key (see the review summary). Under the current design the minimal patch is 2 lines: after `:529` re-read `get_number_by_hash(block_hash)` and compare — treat a mismatch as a miss. That narrows the window to the same order as `insert_by_hash`'s.

Alternative worth considering: if you believe the re-key path is in fact **unreachable** in production, then `Rekeyed` / `Rejected` / `is_canonical` are all dead logic, and the right move is an `assert!` that pins the invariant and fails loud — rather than silently re-keying and adding a layer to absorb it.

Caveat: I derived this from the statement ordering and did **not** reproduce it. If it's worth pinning, a `#[cfg(test)]` yield point between `:554` and `:555` turns it into a deterministic regression test.

bin/debug-trace-server/src/data_provider.rs
```diff
@@ -314,6 +318,42 @@ impl DataProvider {
         }
     }
 
+    /// Whether `hash` is the canonical hash for `number`, used to validate number-keyed
+    /// response-cache hits.
+    ///
+    /// `Some(true)` = canonical: the local DB agrees, or the number is below the DB's
+    /// retained range, where chain-sync reorg handling covered the height before it was
+    /// pruned. `Some(false)` = verified mismatch (the cached entry belongs to a dead
```
vincent-k2026: **Non-blocking — the stated justification for depth-trust doesn't hold for part of the range it covers, and there is no minimum-depth bound.**

The comment grounds `Some(true)` in "chain-sync reorg handling covered the height before it was pruned", but the branch at `:343-345` (number absent from the DB and `<= tip`) also covers **heights that chain sync never stored at all** — everything below the anchor when the node started syncing mid-chain. `on_reorg` was never called for those. The real basis is **finality depth**, not chain-sync coverage.

A reachable fail-open: an entry is cached for a height that was **above the sync tip** at request time (served via RPC), and its hash later turns out to be non-canonical because the reorg happened before chain sync reached that height — so `hashA` was never in the DB and `on_reorg` never fires for it. As long as that entry is not touched while the DB still retains the height (a touch would hit the `Some(false)` mismatch at `:336` and invalidate), pruning it makes this branch approve a dead hash. Narrow, but it fails open rather than closed.

**Suggested fix:** bound the trusted zone to a reorg-safe distance (only trust `number <= tip - REORG_DEPTH`, verify the shallow window upstream) so the trusted range matches the finality assumption you actually rely on, and rewrite the comment to state that reason. Better still, per the review summary: keep `CANONICAL_CHAIN` and prune only bodies/witnesses, and this branch disappears entirely.

bin/debug-trace-server/src/data_provider.rs
```diff
@@ -314,6 +318,42 @@ impl DataProvider {
...
+                    let above_tip = match db.get_canonical_tip() {
+                        Ok(Some(tip)) => number > tip.block_number,
+                        Ok(None) => true,
+                        Err(_) => return None,
+                    };
+                    if !above_tip {
+                        return Some(true);
+                    }
+                }
+                Err(_) => return None,
```
vincent-k2026: **Non-blocking — three `Err(_) => return None` arms (`:341`, `:347`, `:353`) swallow DB and RPC errors with no log and no metric.**

A corrupt DB or a persistently failing endpoint surfaces only as "hit rate mysteriously went to zero", with nothing to attribute it to. Please classify these at least once (`warn!`/`debug!` plus a counter) instead of collapsing them into the same `None` that means "legitimately unverifiable".

bin/debug-trace-server/src/data_provider.rs
```diff
@@ -314,6 +318,42 @@ impl DataProvider {
...
+            }
+        }
+        let deadline = Instant::now() + Duration::from_secs(CANONICAL_CHECK_TIMEOUT_SECS);
```
vincent-k2026: **Non-blocking — the new gate on the hit path has no observability, no coalescing, and a fixed deadline unrelated to the request's own budget.**

1. The tri-state result of `is_canonical` is **not measured anywhere**. `debug_getCacheStatus` (`rpc_service.rs:661-681`) still only reports hits/misses/hitRate, so in production you cannot distinguish "the cache is useless" from "canonicality is unverifiable". This PR added a metric for request shapes but none for the gate it just put in front of every number-keyed hit.
2. `CANONICAL_CHECK_TIMEOUT_SECS = 2` (`:141`) is a hard-coded constant independent of the request's remaining budget. When upstream degrades, every hit burns 2 s and *then* falls back to a full recompute — the cache's value evaporates exactly when upstream is under stress, with no circuit breaker.
3. This upstream lookup is **not single-flighted** (block/witness fetches are), so N concurrent requests for the same block issue N canonical checks.

**Suggested fix:** add an `{outcome=canonical|reorged|unverifiable}` counter and surface it in `debug_getCacheStatus`; add a small `number -> (hash, Instant)` memo in `DataProvider` with a ~500 ms–1 s TTL so a burst collapses into one upstream resolution, and treat `number <= head - REORG_DEPTH` as final (bounded LRU, no expiry).

Also: `Option<bool>` as a tri-state makes the call site (`Some(true)` / `Some(false)` / `None`) unreadable — `enum Canonicality { Canonical, Reorged, Unknown }` would state the rule.

(The latency/QPS half of this only applies in stateless mode; see my question about the deployed mode in the review summary. The observability half applies either way.)

bin/debug-trace-server/src/response_cache.rs
```diff
@@ -131,28 +111,81 @@ pub enum ResponseVariant {
...
+
+        Some(Self::Tracer(tracer_type, config_hash))
+    }
 
-        Self::Tracer(tracer_type, config_hash)
+    /// Coarse request-shape label for observability, classifying like
+    /// [`Self::cacheable_from_geth_options`]: which tracer the caller asked for, with each
+    /// non-cacheable shape kept distinct so its production frequency is measurable.
+    /// Every returned label is listed in [`crate::metrics::REQUEST_SHAPES`].
+    pub fn shape_label(opts: &GethDebugTracingOptions) -> &'static str {
```
vincent-k2026: **Non-blocking — `shape_label` and `cacheable_from_geth_options` are two parallel matches over the same enum, and nothing structurally keeps them in lockstep.**

`shape_label_matches_cacheability` enumerates 9 hand-written cases, so a new alloy tracer variant will **not** trip it; `#[allow(unreachable_patterns)]` below switches off the one thing the compiler could have done here; and the consistency check only covers the "label ∈ `REQUEST_SHAPES`" direction.

**Suggested fix:** a single `classify(opts) -> Shape`, with `Shape::label()` and `Shape::cacheable_variant()` derived from it. Then the two can't drift, and the test pins a classification instead of a case list.

bin/debug-trace-server/src/response_cache.rs
```diff
@@ -248,11 +281,53 @@ impl SecondaryIndices {
...
+    /// (e.g. stateless mode, where no chain sync runs), every response cached for that number
+    /// belongs to a dead block, and the dead keys are returned for eviction. By-hash inserts
+    /// pass `false`: an explicitly requested orphan hash must not redefine the number's
+    /// canonical mapping, so the insert is rejected and nothing may be cached.
+    fn insert(
+        &self,
+        block_hash: B256,
+        block_number: u64,
+        key: ResponseCacheKey,
+        allow_rekey: bool,
```
vincent-k2026: **Non-blocking — a `bool` policy parameter plus three `unreachable!()` arms encodes a structural invariant in comments.**

`:558`, `:591` and `:643` all guard "impossible" combinations with `unreachable!()`, i.e. a panic inside an RPC server, to protect an invariant the type system could enforce instead.

**Suggested fix:** split into `insert_canonical(...) -> Option<HashSet<ResponseCacheKey>>` and `try_insert_for_hash(...) -> bool`. The bool argument and all three `unreachable!()`s disappear together. (Natural next step after @Troublor's `IndexInsert` naming suggestion, which is already applied.)

bin/debug-trace-server/src/response_cache.rs
```diff
@@ -435,10 +552,53 @@ impl ResponseCache {
...
+                    block_hash = %block_hash,
+                    "Skipped caching by-hash response for a non-canonical hash"
+                );
+            }
+        }
+    }
+
     /// Gets a cached response or computes it, coalescing concurrent requests for the same key.
     ///
     /// This is the primary entry point for cache usage. If a response is already cached,
```
vincent-k2026: **Non-blocking — this doc line says "primary entry point", but `get_or_compute` is dead code.**

`:607` is `#[allow(dead_code)] // May be used in future` and the only callers are tests, yet this PR added re-key handling plus a new async test (`:1068`) to that dead path.

The consequence worth noting: `DataProvider`'s single-flight only covers block/witness **fetching** — trace computation and serialization are not coalesced anywhere, so N concurrent identical `debug_traceBlockByNumber` calls each run a full EVM replay. The new "unverifiable → miss" behaviour amplifies that.

**Suggested fix:** at minimum fix the doc line. Then either wire it into the live path (and get coalescing for free) or stop adding logic to it. Not asking to delete pre-existing dead code in this PR.

bin/debug-trace-server/src/rpc_service.rs
```diff
@@ -835,13 +950,130 @@ mod tests {
...
+            Instant::now(),
+        );
+        assert!(by_hash.is_none());
+    }
+
+    /// Number-keyed hits serve only while the cached hash is canonical: a verified mismatch
+    /// invalidates the dead entries and reports a miss; an unverifiable check (stateless
+    /// mode, upstream down) reports a miss without invalidating.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn number_hit_requires_canonical_hash() {
```
vincent-k2026: **Non-blocking — two concrete problems with the new tests.**

1. **The production-shaped path has no positive coverage.** Every RPC-backed case only exercises "upstream hangs → `None`". `Some(true)` / `Some(false)` *via RPC* — the only branch `is_canonical` can take in stateless mode — is untested. There is already a reusable pattern: `crates/stateless-common/src/rpc_client.rs`'s `start_block_number_rpc(...)` is an in-process JSON-RPC responder; exporting it as test-support would cover this.
2. **They wait on real network timeouts.** Measured on `d33bebf`: `is_canonical_unverifiable_without_chain_source` **4.010 s**, `number_hit_requires_canonical_hash` **2.008 s**, while the other 88 tests total ~1.5 s — these two are roughly three quarters of the suite's wall time. `rpc_client.rs:2045-2049` already shows the better convention (drop `per_attempt_timeout` to 150 ms). Making `CANONICAL_CHECK_TIMEOUT_SECS` injectable would do it.

Also `std::mem::forget(listener)` here (`:913`) and in `data_provider.rs:1123` leaks a socket per fixture; the existing convention elsewhere in the repo is `let _listener = listener;`.

bin/debug-trace-server/src/main.rs
```diff
@@ -459,7 +459,9 @@ async fn main() -> Result<()> {
     let chain_spec = load_chain_spec(&args)?;
 
     let response_cache = if response_cache_disabled {
-        info!("Response cache disabled (estimated_items = 0)");
+        warn!(
+            "Response cache disabled (estimated_items = 0); every trace response will be recomputed"
```
vincent-k2026: **Non-blocking — `estimated_items = 0` as the off-switch is the actual root cause; `warn!` only makes the log prettier.**

Production has already demonstrated that this line gets ignored for months (per the PR description). Failing fast means rejecting `0` at arg-parse time and requiring an explicit `--response-cache-disabled` (plus its env var) to turn the cache off.

AGENTS.md
```diff
@@ -107,6 +107,7 @@ Two operating modes:
 - **Local cache mode** — With `data_dir`, enables chain sync to pre-fetch blocks into `ValidatorDB` for faster serving.
 
 The server includes an HTTP response cache (`quick_cache`) for pre-serialized JSON and a `DataProvider` with single-flight request coalescing.
+The response cache only stores idempotent request shapes: the five built-in tracers (keyed by a hash of their `tracerConfig`) and the bare default struct-logger request; JS tracers, `muxTracer`, and struct-logger requests with non-default flags bypass it entirely.
```
vincent-k2026: **Non-blocking — doc-surface is incomplete.**

This documents the whitelist but not that number-keyed hits now require a canonicality check that, in stateless mode, goes upstream. That's a new behavioural contract *and* a new external dependency — arguably more operationally relevant than the whitelist itself, especially together with the failure mode "when upstream is slow, every hit waits up to 2 s and then falls back to a full recompute". Same for the `README.md` section added in this PR.

bin/debug-trace-server/src/rpc_service.rs
```diff
@@ -328,16 +358,17 @@ fn check_cache_by_number(
...
     resource: CachedResource,
     block_hash: B256,
-    variant: &ResponseVariant,
+    variant: Option<&ResponseVariant>,
     method_name: &'static str,
     start: Instant,
 ) -> Option<serde_json::Value> {
     let (cached_value, block_num) =
-        cache.as_ref()?.get_by_hash(resource, block_hash, variant.clone())?;
+        cache.as_ref()?.get_by_hash(resource, block_hash, variant?.clone())?;
```
vincent-k2026: Nit — `variant?.clone()` inline here, while the by-number helper does `let variant = variant?;` first. Two sibling helpers changed in the same PR now read differently, and this expression carries two `?` with different meanings (`cache.as_ref()?` vs `variant?`). Hoisting it would match the other side.

bin/debug-trace-server/src/response_cache.rs
```diff
@@ -495,6 +659,29 @@ impl ResponseCache {
         }
     }
 
+    /// Evicts cache entries whose index entries were purged by a re-keying
+    /// [`SecondaryIndices::insert`], skipping `keep` — on a same-variant re-key the stale set
+    /// contains a key equal to the one just written, whose slot already holds the fresh
+    /// response. Keys re-registered in the index since the purge (a concurrent insert for the
+    /// new hash) are also skipped so their fresh values survive; the unguarded window between
+    /// that insert's cache write and index write remains, costing at worst one recomputation.
+    fn evict_stale_keys(&self, stale: HashSet<ResponseCacheKey>, keep: &ResponseCacheKey) {
```
vincent-k2026: Nit — `&key != keep` is redundant with `!contains_key(&key)`: `keep` was just written back into the index inside the same write lock in `SecondaryIndices::insert`, so `contains_key(keep)` is true. One condition plus a one-line comment would do what the current five lines of doc explain.

vincent-k2026 (changes requested):
## Review

**In one line:** the whitelist + shape-metric half is ready to merge; the canonicality half needs a direction decision first — these 825 lines are almost entirely scaffolding around one thing, namely that the block hash is not in the cache key. Separately there is a door the whitelist leaves open (unbounded key space) that I think should be closed in this PR.

### What's genuinely good

1. The whitelist is the right shape and **fail-closed**: `_ => return None` means a future alloy built-in tracer is not cached by default, which is an order of magnitude safer than the old "fall back to the `Default` variant".
2. Splitting accounting into `get_with_hash` + `record_lookup_outcome` is a real fix rather than a patch — the caller decides whether an entry may be served, so hitRate doesn't lie during a reorg.
3. Using the shape distribution of 1.57M production requests to argue the whitelist is a no-op for production traffic, then adding a shape metric to turn that log-level inference into a first-hand signal — that's the argument structure I'd like to see more often.
4. I checked every claim in the PR description and **they all hold**: `opts.config` is only read by the struct logger (`tracing_executor.rs:88 / 295 / 417`; built-in dispatch destructures `{tracer, tracer_config, ..}`); `GethDebugTracingOptions` has exactly 4 fields and `timeout` is read nowhere in the repo, so field coverage of the key is complete; the chain_sync reorg-invalidation path really exists (`chain_sync.rs:152-169`).

### Blocking

- `response_cache.rs:145-154` — unbounded, caller-controlled cache key space + 64-bit collision.
- `response_cache.rs:554-555` — the cache slot is written before the index is re-keyed, so `get_by_hash` can serve another hash's trace inside that window.

### Design direction (a suggestion; could be the next PR)

Put the hash in the cache key. A key of `(resource, number, variant)` means the slot is **shared by every hash that ever occupied that height**, and every mechanism this PR adds — `is_canonical`, refuse-to-rekey, stale purging, depth-trust, both race windows — is a symptom of that one fact. With `(resource, block_hash, variant)`: by-hash reads and writes directly, structurally unable to poison anything; by-number resolves number→canonical-hash first and then looks up the key, so a reorged height simply resolves to the new hash and misses cleanly.

I initially thought the cost was identical, and on review **one cell isn't**: `prune_history` prunes `CANONICAL_CHAIN` (number→hash) together with the bodies (`server_db.rs:143` plus the orphan-cleanup loop at `:149-162`). So in local-cache mode, a hit on an already-pruned height costs 0 RPCs today via depth-trust, whereas hash-keying would cost 1.

That actually points at a more fundamental fix: **keep `CANONICAL_CHAIN` and prune only bodies and witnesses** (it's the cheapest table in the DB — number→32B). Then hash-keying costs zero RPCs throughout in local-cache mode, and the depth-trust branch in `is_canonical` **no longer needs to exist** — the fail-open I flagged at `data_provider.rs:324-326` disappears with it. The hash-in-key change and that fail-open collapse into one small storage-layer change.

Given this cache has **never been enabled in production** (per the description, it was silently off for months via `ESTIMATED_ITEMS=0`), changing it now carries no migration cost and is a net deletion of code; doing it after it ships and grows dependents only gets more expensive. If you want to keep the scope tight: land the whitelist + shape metric here (independent, clean, mergeable as-is) and move the canonicality half, together with hash-in-key, to the next PR.

### One thing I need confirmed

The latency/QPS cost of `is_canonical` only applies in **stateless mode** (local-cache goes to the DB and is essentially free). Which mode do fra and ore actually run? If it's local-cache, the latency and upstream-load part of my comment on `data_provider.rs:350` can be downgraded wholesale — the missing-observability part stands either way.

### Suggested landing order

1. The `response_cache.rs:145` item — one change removes the collision, the unbounded key space and the hit-rate splitting at once. **I'd do this in this PR.**
2. Keep `CANONICAL_CHAIN`, prune bodies only.
3. Hash in the key (after 2 it is pure deletion, and it also removes the `:554` item and the atomicity residual noted in the description).
4. If you decide against 3: at minimum add the read-side re-check from the `:554` item (2 lines).

### Remaining nits (not worth separate inline threads)

- `ResponseVariant::Default` and `Tracer(TracerType::Default, None)` are two ways to say "default"; only `resource` keeps them from colliding.
- `insert_cache` / `insert_cache_by_hash` differ only in which method they call — they could be one function with a policy argument.
- After `Some(false)`, `check_cache_by_number` calls `invalidate_blocks` → `remove_block(number)`, which also wipes entries a concurrent request just inserted for the new canonical hash (the mirror of Codex's `:673`). Cache churn only, never a wrong answer.
- Pre-existing, just noting: `TracerType::parse` (`response_cache.rs:91`) is used only by tests — this PR removed its sibling `from_geth_options` and left it behind; and `test_tracer_type_parse` is a subset of `test_tracer_type_all_variants`.
- `tracing_executor.rs:467 / 478 / 489 / 739 / 751 / 763` swallow `tracer_config` parse errors via `unwrap_or_default()`: `{"tracer":"prestateTracer","tracerConfig":{"diffMode":"true"}}` (wrong type) falls back to `Default`, so a diffMode request **gets non-diff output with HTTP 200**, where geth would error. Pre-existing and untouched here, so not a requirement for this PR — but it's the same disease as this PR's subject (serving a response that doesn't answer the request) and it lives in the same code path as the first blocking item, so fixing both together is nearly free.

### How I verified

`cargo nextest run -p debug-trace-server` @ `d33bebf`: 88 tests, 85 passed (the 3 failures were my own SIGTERM; they pass when re-run individually), and all 10 relevant new tests green. Separately I checked alloy's `GethDebugTracingOptions` / `CallConfig` / `PreStateConfig` / `FlatCallConfig` definitions, the `serde_json` feature tree (`preserve_order` is indeed enabled), and the `prune_history` and `chain_sync` invalidation paths.

Happy to be corrected on any of this — especially the `:554` item, which I derived from statement ordering and did not reproduce.

vincent-k2026:
Follow-up on the design question from my review — after reading #162, that PR already answers my main cost objection to putting the hash in the response-cache key.

#162's `BlockDataCache` is **keyed by block hash**, and its own rationale puts it better than I did:

> keyed by block hash: entries are immutable facts about that hash, and by-number lookups resolve number to hash against the DB or upstream first (`data_provider.rs:371`), so canonicality is never cached and no reorg invalidation is needed.

Two things follow:

1. Two caches one layer apart in the same binary now use opposite designs. Everything the response cache has to carry — `is_canonical`, `insert_by_hash`, `IndexInsert::Rejected`, `evict_stale_keys`, the depth-trust branch, and the two documented TOCTOU windows — exists *solely* because its key omits the hash. The block-data cache needs none of it, for exactly the reason quoted above.
2. My only real cost objection to hash-keying the response cache — "by-number would then have to resolve number→hash first" — is already paid for in #162 (`data_provider.rs:371`). Once #162 lands, hash-keying the response cache is largely a matter of reusing that resolution, and it turns into a net deletion.

So I'd sharpen my earlier suggestion: instead of "maybe a follow-up", the response cache should end up under the same rule as the block-data cache, and #162 landing makes that cheap. Not asking to widen this PR — the whitelist + shape-metric half can land as-is; this is about where the canonicality half is headed.

crates/stateless-db/src/helpers.rs
```diff
@@ -132,12 +240,7 @@ pub fn write_reset_to_anchor(database: &Database, anchor: &BlockMeta) -> StoreRe
...
         chain.retain(|_, _| false).store_err()?;
-        chain
-            .insert(
-                anchor.block_number,
-                (anchor.block_hash.0, anchor.post_state_root.0, anchor.post_withdrawals_root.0),
-            )
-            .store_err()?;
+        insert_chain_row(&mut chain, anchor)?;
     }
     write_txn.commit().store_err()?;
```
vincent-k2026: **Blocking — after a stale reset, this can permanently pin a non-canonical block for by-number requests.**

The gate here is "is it below the floor", with no finality condition. Chain, traced end to end:

1. A stale reset takes its anchor from **the remote's latest block** (`pipeline/mod.rs:191-199`), triggered by "local tip fell more than `threshold` behind".
2. The reset goes through `write_reset_to_anchor_preserving_history`, which removes rows above the anchor and keeps everything below. Since the node fell behind, `old_tip << anchor`, so `(old_tip, anchor)` is a **hole** and `floor = anchor ≈ the remote tip`.
3. A `debug_traceBlockByNumber(anchor - 3)` arrives: DB miss → upstream resolve → `anchor - 3 < floor`, so this function **persists it**.
4. But `anchor - 3` is three blocks below the tip and can still reorg. Reorg detection only covers `[floor..tip] = [anchor..tip]`, so a reorg at `anchor - 3` is **never detected** and `invalidate_blocks` never touches it.
5. The canonical index is never pruned and the floor only moves on another reset, so the wrong number→hash binding — and the response cached under it — is **permanent**. Every later `debug_traceBlockByNumber(anchor - 3)` returns the trace of a reverted block.

`verify_hash = true` doesn't help here: it proves the header is internally consistent, not that it is canonical at that height in the end. `on_stale_reset`'s `invalidate_all()` doesn't help either — the poisoning happens *after* the reset.

Steady state is fine: with no reset the floor falls back to the first block ever synced, so everything below it is ancient and effectively final. The exposed window is specifically the stretch of `(old_tip, anchor)` near `anchor`, after a stale reset.

**Fix:** gate the write-back on finality, not just on the floor — persist only when `block_number + REORG_SAFETY_MARGIN <= tip` (or check it against `finalized` once). Near-tip below-floor heights would then resolve upstream every time, which is rare enough not to matter.

Or adopt the in-memory memo (see `data_provider.rs:331`), where the same finality gate makes this structurally impossible and this function stops existing.

bin/debug-trace-server/src/data_provider.rs
```diff
@@ -314,39 +318,113 @@ impl DataProvider {
...
+    pub(crate) fn fetch_deadline(&self) -> Instant {
+        Instant::now() + self.block_fetch_timeout
+    }
 
-        // Try to get block hash from local database first.
-        if let Some(db) = &self.db &&
-            let Ok(Some(hash)) = db.get_block_hash(block_num)
+    /// Resolves a block number to its canonical block hash.
+    ///
+    /// This is what makes number-keyed requests safe to serve from the hash-keyed response
```
vincent-k2026, line 470: **The ~25-line alternative to the whole write-back / floor / permanent-index complex.**

What the persistence buys is: a by-number request below the retained range skips one `eth_getHeaderByNumber`. A bounded in-memory memo buys the same thing, minus the on-disk invariant, the migration, the trait-contract change, the downgrade trap, and both blocking items in this review:

```rust
// DataProvider field — ~40 MB at 1M entries; number→hash for a final height is immutable,
// so no TTL is needed and a bounded LRU is exactly right.
canonical_hashes: Mutex<LruCache<u64, B256>>,

pub(crate) async fn resolve_canonical_hash(
    &self,
    block_num: u64,
    deadline: Instant,
) -> DataProviderResult<B256> {
    // 1. Synced range: unchanged, a local index read.
    if let Some(db) = &self.db {
        match db.get_block_hash(block_num) {
            Ok(Some(hash)) => { record_canonical_hash_resolution("db", "ok"); return Ok(hash); }
            Ok(None) => record_canonical_hash_resolution("db", "miss"),
            Err(e) => { record_canonical_hash_resolution("db", "error"); warn!(..); }
        }
    }
    // 2. Below the retained range: in-process memo.
    if let Some(hash) = self.canonical_hashes.lock().get(&block_num) {
        record_canonical_hash_resolution("memo", "ok");
        return Ok(*hash);
    }
    // 3. Upstream, then memoise only if the height can no longer reorg.
    let header = self.rpc_client
        .get_header_with_deadline(BlockId::Number(block_num.into()), true, Some(deadline))
        .await?;
    if block_num + REORG_SAFETY_MARGIN <= self.local_tip_hint() {
        self.canonical_hashes.lock().put(block_num, header.hash);
    }
    Ok(header.hash)
}
```

What that deletes relative to the current design: `write_canonical_hash_below_floor`, the history-floor row, `write_ensure_history_floor` and its migration, `write_reset_to_anchor_preserving_history`, the `get_earliest` semantic change and therefore the `DivergenceLookups` contract change in a shared crate, the downgrade trap, most of the sizing question the 241-line bench answers, plus N1 (`helpers.rs:222`) and N2 (`data_provider.rs:396`) — the finality gate makes N1 impossible and there is no DB write left to block on.

What it costs: a restart drops the memo, so the first request per cold height pays one header RTT — once. Given the 93.7% repeat rate quoted in the description, that is a small, bounded cost.

Keeping `CANONICAL_CHAIN` unpruned (the round-1 suggestion, a −17-line change to `prune_history`) is orthogonal and still worth doing on its own: it makes the *retained* range resolve locally forever, without introducing any island.

bin/debug-trace-server/src/data_provider.rs
```diff
@@ -314,39 +318,113 @@ impl DataProvider {
...
+            }
+            Err(e) => {
+                record_canonical_hash_resolution("upstream", "error");
+                return Err(e.into());
+            }
+        };
+        if let Some(db) = &self.db {
+            // Best-effort: the request already has its answer; a failed write-back only
+            // costs the next request another upstream round trip.
+            if let Err(e) = db.record_canonical_hash(header.number, header.hash) {
```
vincent-k2026: **Blocking — the write-back takes redb's single-writer lock inline on the async request path.**

`db.record_canonical_hash(&meta)` runs synchronously here, with no `spawn_blocking`, inside an async fn awaited straight from a jsonrpsee handler (`rpc_service.rs:229`). `Durability::None` keeps *this* commit's fsync off the path, but `begin_write()` still has to acquire the writer lock — and chain sync's `store_block_data` uses the default `Immediate` (this is the only `set_durability` call in the repo, `helpers.rs:233`), i.e. it holds that lock across an fsync of several MB of bodies and witnesses.

The read-transaction pre-check does filter the common at/above-floor case, but **cold below-floor historical requests are precisely the write-back's target**, so they take the lock inline and block the tokio worker thread — not just the task — for however long the in-flight sync commit takes. The bench's ~25 µs is the uncontended number.

**Fix:** the write is best-effort and the answer is already in hand, so take it off the request path entirely — hand it to a bounded mpsc drained by a background writer (ideally chain sync's own writer, which serialises naturally and can batch). `spawn_blocking` is the minimum acceptable version; it stops occupying an async worker but still contends for the lock.

(Moot under the in-memory memo in the `:331` thread.)

crates/stateless-db/src/helpers.rs
```diff
@@ -109,16 +104,129 @@ pub fn write_rollback_chain(database: &Database, to_block: BlockNumber) -> Store
...
+pub fn read_archived_hash(
+    database: &Database,
+    block_number: BlockNumber,
+) -> StoreResult<Option<BlockHash>> {
+    let read_txn = database.begin_read().store_err()?;
+    let archive = read_txn.open_table(HASH_ARCHIVE).store_err()?;
+    Ok(archive.get(block_number).store_err()?.map(|v| BlockHash::from(v.value())))
+}
+
+/// Archives an upstream-resolved canonical hash — the trace server's lazy write-back.
```
vincent-k2026: **Non-blocking — the "`[floor..tip]` is hole-free" invariant is enforced by convention on this side.**

`write_canonical_hash_below_floor` protects its half structurally (a floor re-check inside the write transaction, under redb's single writer). Rollback has no matching guard: nothing here stops a caller passing `to_block < floor`, which would delete the `CANONICAL_CHAIN` row backing the floor while the ANCHOR floor row survives — `get_earliest` would then return a floor with no row behind it, bisection would walk into the hole, and `find_divergence_point` treats a missing hash inside its range as `LocalChainCorrupt`, which is fatal and needs an operator.

It's safe today only because the one caller is clamped: `find_divergence_point`'s lower bound *is* `get_earliest()` (`divergence.rs:91-93`).

**Fix:** a `debug_assert!(to_block >= floor)` (or clamp + `warn!`) turns that coupling from a comment into an executable invariant.

crates/stateless-db/src/helpers.rs
```diff
@@ -109,16 +104,129 @@ pub fn write_rollback_chain(database: &Database, to_block: BlockNumber) -> Store
...
+        let chain = write_txn.open_table(CANONICAL_CHAIN).store_err()?;
+        if !below_chain_window(&chain, block_number)? {
+            // Dropping the uncommitted txn aborts it; nothing is written.
+            return Ok(());
+        }
+        let mut archive = write_txn.open_table(HASH_ARCHIVE).store_err()?;
+        archive.insert(block_number, block_hash.0).store_err()?;
+    }
+    write_txn.commit().store_err()?;
+    Ok(())
```
vincent-k2026: **Non-blocking — two things worth writing down here.**

1. **Contiguity is assumed, not verified.** For a legacy database this takes `chain.first()` as the floor. That is correct as long as every historical write path only removed a bottom prefix (`prune_history`) or a top suffix (`write_rollback_chain`) — which is true of the code I can see, so production databases should be fine. But the assumption is what makes this safe, and it isn't stated: if any legacy database did contain a hole, the floor would land below it and bisection would hit `LocalChainCorrupt`. One sentence in this doc comment would make the dependency explicit.

2. **This change is a one-way upgrade, and the description doesn't say so.** Once this binary has written below-floor rows, rolling back to the previous binary means `get_earliest` returns `chain.first()` again — now the lowest island row — so bisection walks a range full of holes and hard-fails. That belongs in "Behavior changes" next to the `ESTIMATED_ITEMS=0` note, since rollback is a normal operational lever.

bin/debug-trace-server/src/main.rs
```diff
@@ -658,89 +677,117 @@ async fn history_pruner(
...
-            if let Ok(m) = std::fs::metadata(&db_path) {
-                chain_sync_metrics.set_db_size(m.len());
-            }
         }
+    }
 
-        tokio::time::sleep(interval).await;
+    // DB gauges: `db_earliest_block` is the canonical window's start; the
+    // `db_body_earliest_block` body-retention edge can sit above it under size pressure.
+    let earliest = db.get_earliest().ok().flatten().map(|(n, _)| n).unwrap_or(0);
```
vincent-k2026, line 806: **Non-blocking — the comment contradicts the code, and the metric changed meaning in place.**

"`db_earliest_block` is the canonical-index floor (descends while the index grows)" — it doesn't descend. The value comes from `db.get_earliest()` → `read_history_floor`, and `HISTORY_FLOOR_KEY` has exactly two writers in the repo (`helpers.rs:198` materialisation, `helpers.rs:260` reset). The floor is fixed until the next reset; only `chain.first()` would descend, and that isn't what's read here.

Separately: `db_earliest_block` used to mean the body-retention edge and now means the canonical floor, with the old meaning moved to a new gauge. Every existing dashboard and alert keeps working and silently means something else. I'd invert it — give the *new* concept the new name (`db_canonical_earliest_block`) and leave `db_earliest_block` alone — so nothing downstream has to be found and fixed.

bin/debug-trace-server/src/main.rs
```diff
@@ -658,89 +677,117 @@ async fn history_pruner(
...
+                Err(e) => {
+                    warn!(error = %e, "Failed to read DB file size");
+                    break;
+                }
+            };
+
+            if file_size <= db_max_size {
+                break;
+            }
+
```
vincent-k2026, line 765: **Non-blocking — the size-prune ratchet, and how the permanent index feeds it.**

The loop exits when `file_size <= db_max_size`, but the comment two lines above states the reason that can't happen: redb files never shrink on their own, and `compact()` is listed as out of scope in the description. So once the file has crossed `db_max_size`, that condition is unreachable and every maintenance cycle prunes all the way down to `retain_floor` — body retention collapses to `--size-prune-min-retain` (256 blocks) and stays there.

The new flag is the right stop-loss (much better than consuming the entire retention), but the ratchet is worth documenting, because **this PR makes crossing the threshold more likely**: the permanent canonical index is 4.5–5 GiB fully populated by your own bench. Raising prod `DB_MAX_SIZE` to 24 GB addresses that, but the failure mode should be written down.

It also cuts against the write-back's own value: once body retention is at 256 blocks, nearly every request fetches witness and body upstream anyway, and saving one header fetch matters even less.

bin/debug-trace-server/src/data_provider.rs
```diff
@@ -1002,6 +1196,149 @@ mod tests {
...
+            post_state_root: B256::ZERO,
+            post_withdrawals_root: B256::ZERO,
+        };
+        // The anchor-init path installs the canonical window at {100}.
+        ChainStore::reset_to_anchor(&*server_db, &tip).unwrap();
+
+        let provider = provider_with(&url, Some(Arc::clone(&server_db) as Arc<dyn BlockStore>));
+        let deadline = Instant::now() + Duration::from_secs(2);
+        let resolved = provider.resolve_canonical_hash(42, deadline).await.unwrap();
+        assert_eq!(resolved, test_support::consistent_header(42).hash);
```
vincent-k2026: **Non-blocking — good test, but it doesn't cover the window that matters.**

Killing the upstream and proving the second resolve is served locally, plus asserting the floor didn't move, is exactly the right shape — this discriminates.

What it doesn't reach is the risky configuration: floor is 100 and the write-back is at 42, i.e. 58 blocks below the floor, which is the safe case. The problematic one (see `helpers.rs:222`) is a height *just* below the floor right after a stale reset, where the floor sits near the tip. A case with floor = tip and a write-back at `tip - 3` would fail today and pass once the finality gate is added.

bin/debug-trace-server/src/response_cache.rs
```diff
@@ -395,120 +476,37 @@ impl ResponseCache {
...
     ) {
-        let key = ResponseCacheKey::new(resource, block_number, variant);
-        let cached = CachedResponse::new(response);
-
-        self.inner.cache.insert(key.clone(), cached);
-        self.inner.indices.insert(block_hash, block_number, key);
+        let key = ResponseCacheKey::new(resource, block_hash, variant);
+        self.inner.index.insert(key);
+        self.inner.cache.insert(key, CachedResponse::new(response));
         self.update_size_metrics();
```
vincent-k2026, line 496: Nit — the doc explains the immediate-eviction direction of this ordering, but not the other one: if a reorg's `invalidate_blocks` lands between `index.insert` and `cache.insert`, it removes the key from the index and finds nothing in the cache, then this insert lands and the entry sits in the cache unreferenced by the index. Harmless now — with the hash in the key, that entry is a true fact about that hash and is unreachable from number-keyed requests — so it just occupies bytes until weight eviction. Worth one clause in the comment, since the current text implies the ordering was chosen with only the eviction race in mind.

vincent-k2026 (changes requested):
## Review (round 2, against `706a29c`)

### All 13 items from round 1 are addressed

Verified one by one, and I want to be specific about the good parts before the asks:

- **Hash in the key** (`75985bd`): `+900 / −1173` across 6 files — a **net deletion of 273 lines**, exactly the shape this was supposed to take. `is_canonical`, `insert_by_hash`, `evict_stale_keys`, `IndexInsert`, `DefaultHasher`, `allow_rekey`: zero references left in the repo.
- **The typed variants are complete**, not approximate: `CallConfigKey` 2/2 fields, `PreStateConfigKey` 3/3, `FlatCallConfigKey` 2/2, so no two configs that produce different responses can collapse onto one key. Key space is closed.
- `invalidate_blocks`'s comment — "pure memory hygiene — a dead hash's entries are unreachable by number-keyed requests" — tells me the post-refactor invariant is actually understood rather than transcribed. Same for `RequestShape`: folding classification, label and cache variant into one three-state enum is cleaner than what I suggested.
- `--response-cache-disabled` plus `value_parser!(u64).range(1..)` rejects `0` on both CLI and env; `classify_and_gate` returns `-32602` before any fetch; `mem::forget` is gone repo-wide; the 2s/4s timeout tests went away with the code they tested.

Independently checked rather than taken on trust: `verify_hash = true` really does recompute `hash_slow` client-side (`rpc_client.rs:530/544`); redb 4.0.0's own docs back the durability claim (`transactions.rs:362-369`); the `get_earliest` semantic change does **not** break the validator (`ValidatorDB::get_earliest` still returns its oldest row, it has no below-floor islands); `write_advance_chain`'s refactor only extracted `insert_chain_row` — the max-len pruning is byte-for-byte unchanged, so the shared-crate blast radius really is as narrow as claimed. `cargo nextest run -p debug-trace-server -p stateless-db` @ `706a29c`: **114 passed, 0 failed**.

### The main ask: about three quarters of this PR is optional, and every new risk lives in that part

Accounting by commit:

| commit | what | net |
|---|---|---|
| `75985bd` | hash-keyed cache + typed variants (**= the entire review response**) | **−273** |
| stop pruning `CANONICAL_CHAIN` | deletes 3 lines + a 14-line orphan-cleanup loop in `prune_history` | **−17** |
| `e310e76` + `54c7218` + `c9a27df` + `706a29c` | floor / backfill (written then deleted) / bench / lazy write-back | **+864** |

So the two review items cost **about −290 lines across 7 files**. The remaining **+864 net lines hang off one optional optimisation: the lazy write-back.** And every new risk in this PR is downstream of it:

```
lazy write-back  (to save one eth_getHeaderByNumber)
 └─ creates non-contiguous "islands" below the synced range
     └─ breaks get_earliest()'s hole-free assumption
         ├─ forces the history-floor concept
         │   ├─ forces the get_earliest() semantic change
         │   │   └─ forces the DivergenceLookups contract change (shared crate, validator)
         │   ├─ forces the write_ensure_history_floor migration
         │   └─ creates a downgrade trap (old binary reads chain.first() → walks into a hole → fatal)
         ├─ creates N1 (post-stale-reset poisoning, below)
         └─ creates N2 (redb write on the async request path, below)
     └─ motivates the 241-line bench
```

**What it buys:** only when a by-number request lands below the retained range *and* hits the response cache does it turn one header RTT into two local reads. On a response-cache miss the same request still fetches body + witness upstream, next to which a header fetch is a rounding error.

**The number that would justify it is missing.** The PR argues from "93.7% repeat-after-success", but never states **what share of by-number requests fall below the retained range**. That share is what decides whether the subsystem earns its keep — and note it interacts badly with the size-prune ratchet (see the `main.rs:731` thread).

**The ~25-line alternative** is sketched in the `data_provider.rs:331` thread: a bounded in-memory `number → hash` memo gated on a finality margin. It keeps essentially all the benefit and deletes the floor, the migration, the trait-contract change, the downgrade trap, the bench's sizing question, and both blocking items below — N1 becomes structurally impossible because a reorg-able height is never memoised.

**Some of this is on me.** Round 1 said "keep `CANONICAL_CHAIN`, prune bodies only", and I added that it would "collapse into one small storage-layer change" — which reads like a licence to build a persistence story. What I meant was *stop deleting rows you already write* (a −17-line change to `prune_history`), not *build a permanent index with a floor and a write-back path*. Sorry for the ambiguity.

### Blocking

- `helpers.rs:222` — **N1**: after a stale reset, a below-floor write-back can permanently pin a non-canonical block for by-number requests.
- `data_provider.rs:396` — **N2**: the write-back takes redb's single-writer lock inline on the async request path.

Both are artefacts of the write-back; both disappear under the alternative.

### Non-blocking

`helpers.rs:154` (rollback has no floor guard) · `helpers.rs:187` (legacy contiguity assumed, not verified; plus the downgrade trap) · `main.rs:782` (comment contradicts the code; metric changed meaning in place) · `main.rs:731` (size-prune ratchet, and how the permanent index accelerates it) · `data_provider.rs:1270` (the write-back test doesn't cover the risky window) · `response_cache.rs:486` (nit).

### Suggested split

| # | what | size | when |
|---|---|---|---|
| 1 | hash-keyed cache + typed variants | net −273 | mergeable now |
| 2 | stop pruning `CANONICAL_CHAIN` + `DB_MAX_SIZE` bump + the bench's sizing conclusion | net −17 + docs | mergeable now |
| 3 | below-retention resolution: start with the in-memory memo (finality-gated) | ~+25 | mergeable now |
| 4 | if measurement shows the memo is insufficient, *then* persistence — its own PR, with floor, migration and downgrade notes | — | needs data |
| 5 | `--size-prune-min-retain` fix, metric rename, the "4-angle quality pass" | — | separately |

Items 1–3 are the whole functional story at roughly one quarter of the current diff, with none of the new on-disk surface.

### How I verified

Read the full diff and the post-rewrite files; traced the stale-reset path end to end (`pipeline/mod.rs:180-199` → `write_reset_to_anchor_preserving_history` → floor → `write_canonical_hash_below_floor`); traced the rollback clamp (`advancer.rs` → `find_divergence_point` → `divergence.rs:91-93`); read redb 4.0.0's `Durability` docs; diffed `helpers.rs` against `d33bebf` to size the validator regression surface; ran the two test packages.

Happy to be wrong on the framing — if the below-retention share turns out to be large, item 4 becomes justified and I'd rather see it as its own PR than not at all.

bin/debug-trace-server/src/data_provider.rs
flyq: Here is the missing number, from the 3.6-day production replay (fra 85.8h + ore 99.4h, `BLOCKS_TO_KEEP=10000`; all by-number traffic is `debug_traceBlockByNumber`+callTracer — by-hash and `trace_block` are literally zero in the logs):

| regime | by-number requests below the retained range |
|---|---|
| fra, full window (backfill storm included) | 1,210,706 / 1,537,105 ≈ **79%** |
| fra, steady state (last 18h) | ≈ **4–5%** (~186/h block-level historical vs ~4,000/h near-tip) |
| ore, 99.4h | ≈ **5%** (51,015 historical executions vs 885,076 DB-served reads) |

At the measured chain rate (~1.6 blocks/s) the 10k-block window is only ~1.5–2 h of chain, so all backfill/rescan traffic is below it by construction.

Two qualifications, one in each direction:

1. You're right that on a cache **miss** the header RTT is a rounding error next to body+witness. But 93.7% of the 1.22M below-retention executions were repeats-after-success — with the response cache now enabled those become **hits**, where number→hash resolution is the *only* upstream call left. That is ~75% of all by-number requests in the window; on the storm day, ~11/s sustained — and resolution is not single-flighted. Storms are not a tail case here: the 07-14 incident, the 07-24 rescan, and the planned full re-run of the 1M-block stress segment are all below-retention workloads.
2. Conceded: steady state the share is 4–5% and your memo covers most of the benefit. The archive's increment over the memo, from the same data: pruning feeds it ~143k locally-verified rows/day at zero RPC (every post-deploy height resolves locally even on first touch), and it survives restarts — a mid-storm restart re-buys ~50k resolutions on a cold memo (07-24 had 49,787 unique heights). Cost is bounded and small: 84–135 B/row, ≈2.3–2.5 GiB fully populated.

On N1/N2: both are fixed in the next revision of this branch, in the shape you suggested — details in the respective threads.

crates/stateless-db/src/helpers.rs
flyq: Fixed in the next revision of this branch, in the shape you suggested — and the floor machinery this thread was written against is gone entirely (`b11f78b` replaced it with a dedicated `HASH_ARCHIVE` table; `CANONICAL_CHAIN` is back to the bounded contiguous window and `get_earliest` is `chain.first()` again, so the trait-contract change, the migration, and the downgrade trap all went away with it).

The same hazard did carry into the archive design (a fresh anchor init or stale reset installs a single-row window at the remote tip, so "below the window" could mean one block deep). The write-back gate is now depth-final in both checks (read pre-check and write-txn re-check): `block_number < window_start && block_number < tip − ARCHIVE_MIN_DEPTH` (64). Your exact scenario — floor = tip, write-back at tip−3 — is pinned by a test (`record_canonical_hash_only_archives_depth_final_heights`: skipped at tip−1 and at the boundary, archived one below it).

Relatedly (Codex found the sibling case): pruning now archives only the `CANONICAL_CHAIN` row's hash — `BLOCK_RECORDS` reorg siblings at the same height can no longer feed the archive — with a regression test that puts dead hashes on both byte-order sides of the canonical one.

bin/debug-trace-server/src/data_provider.rs
flyq: Fixed in the next revision of this branch: the write-back now runs fire-and-forget via `spawn_blocking` — the request returns as soon as the upstream answer is in hand and never waits on the writer lock; a lost write costs one refetch, and the e2e test polls for the async landing. Went with your "minimum acceptable version" deliberately: the write is gated to depth-final heights and idempotent, so blocking-pool contention is bounded by cold-first-touch volume. If storm-time telemetry shows the lock contention mattering, the bounded-mpsc-into-the-sync-writer upgrade (with batching) is the natural next step.

bin/debug-trace-server/src/data_provider.rs
flyq: Adopted — the branch now carries your memo instead of the persistent archive (`242bbb9`, following `b11f78b`'s interim `HASH_ARCHIVE` design which this supersedes). `resolve_canonical_hash` goes window → bounded in-memory LRU → upstream, memoizing only when `block_num < observed_tip − 64`; the tip reference is a monotonic max of verified on-chain heights (latest-tag resolutions, verified headers, the local window tip), which can only under-estimate the real tip, so the gate errs conservative and works in stateless mode too. Capacity is CLI-tunable (`--canonical-hash-memo-capacity`, default 8M entries ≈ 0.6 GB if ever full, filled lazily — the measured working set is ~50–130k heights and our boxes have 128 GB).

What decided it, pricing the archive's increments over the memo against the replay: restart rewarm is ~200 RTTs/h at steady state (a storm rewarm spreads over the storm's own pace), and the pruning feed mostly populates heights nobody ever queries — while the archive's costs (permanent blast radius for any feed-path bug, monotonic disk growth into the size-prune ratchet, an invariant spread across prune/reset/write-back) are the kind that bite later. Your N1 and N2 are now structurally impossible rather than fixed: a reorg-able height is never memoized, and no DB write exists on the request path. Net vs the archive revision: −163 lines and zero new on-disk surface; `canonical_hash_resolution_total{source="memo",outcome="miss"}` is exactly the telemetry that would justify your item 4 (persistence, its own PR) if the miss volume ever matters.

One deliberate deviation from your split: `prune_history` still prunes chain rows (your row 2, the −17 change, is not taken). Unpruned chain rows grow disk at ~3× the archive's rate (96 B values — your own 4.5–5 GiB full-meta figure), evaporate on every stale reset anyway, and the memo covers their benefit; your ratchet objection argues against them too.

crates/stateless-db/src/helpers.rs
flyq: Superseded by the memo revision (`242bbb9`): the write-back path this thread is about no longer exists — a reorg-able height is never memoized (`block_num < observed_tip − 64` on a tip hint that never exceeds the real tip), and a memo entry cannot outlive the process, so the poisoning is structurally impossible rather than gated. Details in the `data_provider.rs:331` thread.

bin/debug-trace-server/src/data_provider.rs
flyq: Superseded by the memo revision (`242bbb9`): there is no DB write on the request path anymore — the memo insert is an in-memory LRU put. Details in the `data_provider.rs:331` thread.

bin/debug-trace-server/src/chain_sync.rs
```diff
@@ -180,83 +180,51 @@ impl PipelineHooks for TraceHooks {
 mod tests {
```
vincent-k2026, line 198: **Non-blocking — the memo has no invalidation hook.**

`on_reorg` and `on_stale_reset` clear the response cache but never touch `canonical_hash_memo`. Two regimes:

- **local-cache mode (what fra/ore run): already safe.** `resolve_canonical_hash` consults the DB first, and the window is 10k blocks deep, so any height a reorg can touch is answered by the DB — which chain sync corrects. The memo only ever serves heights *below* the window, where a reorg is not a thing.
- **stateless mode: no correction path at all.** No DB, no chain sync, no `on_reorg`. The 64-block depth gate is the only defence, and a reorg deeper than that would pin the wrong hash for those heights permanently, with nothing to detect it.

A `memo.clear()` alongside the existing `invalidate_all()` here is one line and closes the local-cache path belt-and-braces. For stateless, the assumption ("reorgs never exceed `CANONICAL_MEMO_MIN_DEPTH`") is currently implicit — worth stating next to the constant, since it is the whole safety argument there.

bin/debug-trace-server/src/data_provider.rs
```diff
@@ -311,42 +350,119 @@ impl DataProvider {
...
+        };
 
-        // Fall back to RPC. The deadline carries through to `get_block_data_by_hash_inner` so
-        // there is only ONE budget shared across hash resolution + the full pipeline.
-        let block_hash =
-            self.rpc_client.get_block_hash_with_deadline(block_num, Some(deadline)).await?;
-        self.get_block_data_by_hash_inner(block_hash, deadline).await
+        // Memoize only depth-final heights. The tip hint never exceeds the real tip, so a
+        // gate that passes on it is final — consult the sync window's tip only when the
+        // hint alone cannot admit the height (typically just the first resolution after
```
vincent-k2026: **Non-blocking, but this is the load-bearing invariant, so it's worth stating precisely.**

"The tip hint never exceeds the real tip" isn't strictly true: `observe_tip` is a `fetch_max`, and the real tip *decreases* during a reorg. After one, the hint sits above the current tip by the reorg depth.

The safety conclusion still holds — the deviation is bounded by reorg depth, which the 64-block margin absorbs comfortably — but the accurate statement is something like: *the hint never exceeds the highest height ever seen as canonical, so the effective depth margin is `CANONICAL_MEMO_MIN_DEPTH` minus the deepest reorg since startup.* Since this sentence is the entire reason N1 cannot come back, I'd rather it said what it actually guarantees.

vincent-k2026 (approved):
## Approving — verified against `cdaedda`

Both blocking items are resolved **structurally rather than patched**, which is the outcome I was hoping for:

- **N1** (a reorg-able height being pinned forever): a height is memoised only when `block_num < tip_hint − 64`, and a memo entry cannot outlive the process. There is no gate to get wrong at runtime.
- **N2** (a redb write on the async request path): there is no DB write on that path at all any more — it's an in-memory LRU put.

And the design arrived somewhere smaller than where it started. Net of the whole branch: **+245 lines of production code**, +146 comments, +389 tests. `response_cache.rs` production code went 436 → 418; `tracing_executor.rs` is untouched; the two "cleanup"/"simplify" commits were **−88** and **+4** production lines respectively — they shrank the branch rather than padding it. `HASH_ARCHIVE`, the history floor, the migration, the trait-contract change and the downgrade trap are all gone, and `record_canonical_hash` leaves no dead interface behind.

### Checked this round rather than taken on trust

- **The N1 regression test discriminates properly.** Height 42 (deep) is memoised, height 199 (tip 200, one block deep) is not — and it proves it by killing the upstream: 42 still serves from the memo, 199 now fails. That is the "remove the fix and the test goes red" standard.
- `--canonical-hash-memo-capacity` ships with `value_parser!(u64).range(1..)` — the `ESTIMATED_ITEMS=0` lesson applied to a new flag without being asked. Thank you.
- I suspected an 8M default would preallocate ~0.6 GB at startup, so I read quick_cache 0.6.16: `HashTable::with_capacity(0)` / `LinkedSlab::with_capacity(0)`. "Filled lazily" is accurate.
- `get_block_hash_with_deadline`, removed from `stateless-common` in the simplify pass, has zero remaining references — dead code removed at the source rather than one link short.
- Everything from round 2 is either gone with the floor (rollback lower bound, legacy contiguity, downgrade trap, contract change, the write-back test window) or fixed: `get_earliest` is `read_earliest_block` again with a comment on why the window is contiguous by construction; the `db_earliest_block` comment no longer contradicts the code and the two gauges now move together, which defuses my naming objection; the size-prune ratchet is documented on the flag.
- `debug_trace_request_shape_total` and `debug_trace_canonical_hash_resolution_total{source,outcome}` landed with exactly the label sets that make the memo's value measurable.
- `cargo nextest run -p debug-trace-server -p stateless-db` @ `cdaedda`: **102 passed, 0 failed**.

### Two corrections on my side

1. **I measured the wrong thing in round 2.** The `−273 / +864` table was line-level diff, and I let it read as "code volume". Counted by content, the branch is +245 production lines and `response_cache.rs` actually shrank. The table's *conclusion* stands (it was a same-unit comparison and the optional part really was the bulk), but the absolute numbers were not code size, and the two cleanup commits I worried about were net deletions. Sorry for the noise.
2. **You were right to reject "stop pruning `CANONICAL_CHAIN`".** Unpruned chain rows grow disk faster than the archive they'd replace, evaporate on every stale reset, and the memo covers their benefit — and, as you pointed out, my own size-prune ratchet objection argues against keeping them. Dropping that item.

### Non-blocking follow-ups (not gating this merge)

1. `chain_sync.rs` — the memo has no invalidation hook; one line would close the local-cache path completely. Details inline.
2. `data_provider.rs:434` — the comment states the tip hint "never exceeds the real tip", which isn't strictly true of a monotonic max. Details inline.
3. **Pre-existing, just noting:** when the sync pipeline hits a `Fatal` (e.g. `CatastrophicReorg`) it halts, and `main.rs` only logs "Chain sync pipeline exited with error" — the process keeps answering RPC from a local window that is frozen and known-divergent, with no health signal. Not introduced here (those lines aren't in this diff), but this PR makes that window the first tier of by-number resolution and the reference the memo's depth gate leans on, so it carries more weight than it used to. Worth an issue.

Nice iteration — five revisions in a day, each one smaller than the last, with the production numbers to back the decisions.

vincent-k2026 (approved):
Approving the post-approval delta (`efe0023` + the `main` merge `4a09369`). Re-reviewed round 4 against the current head, plus a fresh adversarial pass over the hash-keyed key and the memo; `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features` (0 warnings) and `cargo test --workspace` all green on `4a09369`.

**Merge is clean.** `4a09369`'s entire footprint is `.github/workflows/claude.yml` — not one business line touched, so there was no semantic conflict to resolve.

**All three round-4 items close, with discriminating tests.**

- *Tip seed (Codex, `data_provider.rs:442`)* — `seed_tip_from_upstream` (`:438-458`), 30 s-throttled and `TIP_SEED_TIMEOUT`-capped. `resolve_canonical_hash_seeds_tip_for_stateless_numeric_traffic` fails on the unfixed code (`tip_hint` stays 0, nothing memoizes, the post-`handle.stop()` resolve can't be served), and `seed_hits == 1` pins the throttle. The pre-existing depth-final test gained a `seed_hits == 0` positive control, which is a nice negative control on the new path.
- *Memo invalidation* — `on_reorg` → `memo.on_reorg(depth)` (`chain_sync.rs:180`) and `on_stale_reset` → `memo.clear()` (`:192`); `only_margin_deep_reorgs_clear_the_memo` pins the exact `64−1` keep / `64` clear boundary, and the stateless assumption is now stated on the constant.
- *Tip-hint invariant* — restated at `data_provider.rs:336-343` in the requested shape, and the code does satisfy the weaker guarantee it now claims.

I also re-checked the claims the redesign rests on and they hold: the hash really is in the key on every path (by-number resolves the hash *before* the lookup at `rpc_service.rs:226-234` and inserts under the fetched hash at `:503`); the variant space is a closed 48 values, and alloy's `GethDebugTracingOptions` v1.1.2 carries only `config`/`tracer`/`tracer_config`/`timeout` — no `stateOverrides`/`blockOverrides` (those live on `GethDebugTracingCallOptions`, which these methods don't take), so there is no override-vs-cached-non-override collision to worry about; upstream headers are verified on **both** number (`rpc_client.rs:1182`) and hash (`:1201`); and the shared-crate delta is `open_read_table` getting strictly more lenient plus a doc edit, so "no new on-disk surface, zero migration, downgrade-safe" checks out for the validator binary and an external embedder alike.

Two follow-ups, neither blocking.

**1. The `classify_and_gate` wiring has no test protecting it.** `rpc_service.rs:575` is `let _ = classify_and_gate(...)`. The gate is unit-tested, but delete that line from any of the three handlers and the entire suite still passes — and that line is the `-32602` boundary this PR just built. One handler-level test per opts-taking method (`{"tracer":"callTracer","tracerConfig":{"onlyTopCall":"yes"}}` → `-32602`) would close it.

**2. The tip seed is awaited inline.** At `:548` the answer is already in hand (`:531`), yet the request that wins the throttle waits up to `TIP_SEED_TIMEOUT` before returning at `:554`, purely to enable *future* memoization. Spawning it detached would drop that cost; this round simply wouldn't memoize, which the code tolerates everywhere else. Bounded (≤1/30 s), so follow-up-shaped.

Nits: the "deepest reorg since startup" phrasing (`:341-343`, mirrored in the README) is mine and it's slightly loose — the real bound is `max-height-ever-seen − current tip`, which can exceed the deepest *single* reorg when reorgs stack while the tip is below its all-time peak (peak 100 → −10 → 90 → −10 → 80 gives overshoot 20). The load-bearing first clause is exact, so no action needed unless the doc is touched again. Separately, `--size-prune-min-retain` floors only the size-based loop — count-based pruning at `main.rs:735` is unaffected — which matches the doc comment but invites the absolute-floor reading; a line of `help` text would help operators.

Troublor (approved):

PR #162, opened 2026-07-27
feat(debug-trace-server): bounded in-memory BlockData cache keyed by block hash

flyq:
## Summary
Adds a bounded memory tier between the HTTP response cache and the DB / RPC tiers so requests that miss the response cache — other tracer variants of a block, and especially `debug_traceTransaction` calls, which are never response-cached — reuse one witness fetch and contract resolution instead of repeating it per request. Stacked on #161; retarget to `main` after it merges.

## Motivation
3.6 days of production logs show 73% of tx-driven block fetches (9,955 of 13,716) re-fetched a block already resolved moments earlier for a different transaction of the same block, plus ~345k repeated DB reads (redb read + witness decode each) for recent blocks; repeat-same-tx is ~0.4%, so a tx-level response cache would absorb nothing — the reusable unit is the block's resolved `BlockData`.

## Design
New `BlockDataCache` (`bin/debug-trace-server/src/block_data_cache.rs`) backed by `quick_cache`, keyed by block hash: entries are immutable facts about that hash, and by-number lookups resolve number → hash first (via #161's window → memo → upstream chain), so canonicality is never cached and no reorg invalidation is needed. Lookup order is memory → DB → RPC; DB hits are written back, and the RPC-path insert runs inside the shared single-flight future so each fetch inserts exactly once even if the primary caller is cancelled. Memory hits record their block-distance sample from a tip memoized on DB-backed serves, so a pure cache hit opens no redb transaction.

Entry weights are computed once at insert (`block_data_weight`) and stored with the entry, so the weigher quick_cache re-invokes under its shard locks only reads a number; the weight charges the witness KV/level payload (`stateless_common::witness_size::light_witness_memory_bytes`, per-KV constant now `size_of`-derived so a `salt` layout change propagates), contract bytecodes (via the shared `stateless_db::bytecode_weight` formula), and per-transaction envelope overhead plus calldata, access lists, and EIP-7702 authorization lists. The cache pins 4 shards so the admission ceiling for a single entry is `max_bytes / 4` on every host (quick_cache's default shard count scales with the core count, which would shrink the largest admissible block on big machines); startup logs the effective per-shard budget, and an insert the cache does not retain is counted (`cache_admission_rejects_total`) and logged instead of masquerading as a miss.

Trace failures are typed (`TraceError` in `tracing_executor.rs`) and cache hygiene keys off the discriminant: data-attributable failures — a witness that cannot replay the block; `run_transaction` and prestate-frame failures now abort the trace instead of degrading into cacheable `TraceResult::Error` entries — evict the block's entry and are counted per method (`block_data_evictions_total`), while request-attributable failures (mux/JS config parse, inspector construction, tracer output) never evict, so a client cannot drop hot entries with a cheap invalid request. A visible behavior change rides on this: `debug_traceBlock*` on a block whose witness cannot replay now returns `-32000` instead of `200` with per-tx error entries — consistent with parity `trace_block` and geth's abort behavior, and it keeps the response cache from storing poisoned callTracer responses.

New knob `--block-data-cache-max-size` / `DEBUG_TRACE_SERVER_BLOCK_DATA_CACHE_MAX_SIZE` (default 1GB, `"0"` disables). `debug_getCacheStatus` always reports both `responseCache` and `blockDataCache` sections, and the cache exports hits/misses/entries/bytes/admission-rejects under `type="block_data"` plus a `memory` data-source label. Combined default memory across the response, block-data, and contract caches is ~2.5 GB, now stated in the README.

## Testing
Unit tests on real fixture data: weigher terms including exact access-list/authorization arithmetic, hit/miss/stats counters, deterministic single-shard budget eviction (exactly two of three same-weight entries retained — the earlier form passed on a cache that admitted nothing), oversized-entry rejection counting, DB write-back with `Arc::ptr_eq` sharing plus a memoized-tip assertion (the memory-served read performs zero tip reads), memory short-circuit ahead of a hanging upstream, disabled-cache DB re-reads, executor-level Data/Request error discrimination, handler-level evicts-only-on-data-errors, both `debug_getCacheStatus` JSON shapes, and flag default/env/disable parsing. Full suite: 99 bin tests + 39 stateless-common + 17 stateless-db, `cargo fmt --check` / `clippy --workspace --all-targets --all-features` / `cargo sort --check` all clean.

## Notes
Deployment is config-free by default (cache on at 1GB); operators can raise, lower, or disable via the env var. Known follow-up: a deterministically bad witness still costs an evict → refetch → fail cycle per request (visible via the eviction counter); negative caching with backoff is deferred. Merge flow per repo convention: after #161 merges, retarget this PR to `main`, "Update branch" (merge, no rebase), then squash-merge.


bin/debug-trace-server/src/rpc_service.rs
```diff
@@ -488,6 +488,7 @@ impl DebugTraceRpcServer for RpcContext {
         .await
         .inspect_err(|_| {
             metrics::record_rpc_error(METHOD_DEBUG_TRACE_BLOCK_BY_NUMBER);
+            self.data_provider.evict_block_data(&block_hash);
```
vincent-k2026: **Blocking — the signal this eviction keys off ("did the outer `Result` come back `Err`?") is wrong in both directions.**

**It catches none of the cases it was written for.** A decodable-but-incomplete witness shows up as a state read failing inside one transaction. Every block-level tracer turns that into a `TraceResult::Error` element and returns `Ok`: `tracing_executor.rs:328-332` (the shared path for callTracer / prestateTracer / flatCallTracer / default struct logger), `:517` (fourByte), `:582` (mux), `:675` (js). The outer result is `Ok`, `inspect_err` never runs, and the bad entry stays cached. For `debug_traceBlockByNumber` / `ByHash` the commit's stated purpose — "evict block data that fails execution" — never fires.

**It fires on cases that have nothing to do with the cached data.** An invalid mux config or JS tracer fails while *constructing* the inspector, before `setup_executor!`, i.e. before a single byte of EVM state is read: `tracing_executor.rs:533`, `:535`, `:608`. So any client can drop hot blocks from the cache at will with `{"tracer":"muxTracer","tracerConfig":{"bogus":1}}` — and since mux/JS bypass the response cache, every such request really does reach this path. One cheap invalid request forces every other caller back through a witness fetch and contract resolution.

**The behaviour is also inconsistent across entry points.** For the same bad witness: `debug_traceBlockBy*` does not evict (`:328`), `trace_block` (parity) does — it returns `Err(ValidationError::BlockReplayFailed)` on a per-tx failure (`:906`) — and `debug_traceTransaction` does too, via `replay_preceding_txs!` (`:207-210`). Same data, opposite outcome depending on which method you call.

**Worth flagging alongside #161:** with a bad witness, callTracer returns a *successful* response full of `TraceResult::Error` entries. callTracer is on #161's cacheable whitelist and is 97.9% of production traffic, so that response gets stored in the response cache and served indefinitely; the new canonicality check can't help, because the hash really is canonical — the data is what's wrong.

**Suggested fix.** Decide where the error still has a type, not at this call site where it has been erased to `ErrorObjectOwned`:
- Evict only for data-attributable variants — `ValidationError::BlockReplayFailed`, `EnvOracleConstructionFailed`, `BlockIncomplete`. `BlockReplayFailed` already exists, so the rule is expressible today.
- Never evict for tracer-construction errors (invalid mux config, `JsInspector` creation failure, etc.).
- To actually cover bad witnesses on the geth block path, surface a *typed* signal when a per-tx failure comes from a witness/DB read miss instead of stringifying it into `TraceResult::Error`. Ordinary EVM reverts are already `Success`, so they're unaffected.
- Even with correct discrimination: for a *deterministic* bad witness, evict → refetch the same bad witness → fail → evict converts one cheap in-cache failure into repeated expensive upstream fetches. Negative caching with backoff fits better than unconditional removal.

**Testing:** none of the five `evict_block_data` call sites has handler-level coverage — nothing asserts when eviction should or should not happen, which is why both directions survived. A test for the rule is really the acceptance criterion for this fix.

bin/debug-trace-server/src/block_data_cache.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// Creates a new cache bounded to `max_bytes` of estimated memory.
+    pub fn new(max_bytes: u64) -> Self {
+        Self {
+            cache: Cache::with(
+                BLOCK_DATA_CACHE_ESTIMATED_ITEMS,
+                max_bytes,
+                BlockDataWeighter,
+                RandomState::default(),
+                DefaultLifecycle::default(),
+            ),
```
vincent-k2026: **Non-blocking, but please make it observable — the largest cacheable block is a function of the host's CPU count, not of `max_bytes`.**

`Cache::with(..)` goes through `OptionsBuilder` defaults (quick_cache 0.6.16 `sync.rs:86-102`), and the default shard count is `available_parallelism() * 4` (`options.rs:128`). An entry must fit within a *single shard's* slice of the budget or it is silently not admitted — this PR never sets `shards` explicitly.

Measured (binary search over `max_bytes` with one fixture entry of weight 95,589 B, on a 12-core host): **a single entry needs `max_bytes >= 33x its own weight` to be admitted at all.** Extrapolating the default 1 GB:

| host cores | largest cacheable block |
|---|---|
| 12 | ~31 MB |
| 64 | ~6 MB |
| 96 | ~4 MB |

**More cores means a smaller admissible entry**, which is the opposite of the intuition, and a trace server runs on large machines. The README sentence "blocks heavier than a cache shard's budget are never admitted" is literally true, but a reader can't derive that budget: it's `max_bytes / (available_parallelism() * 4)`, further reduced by the hot/cold split.

The part that worries me is that it's unobservable: `insert` (`:165`) is unconditional and a rejected insert looks exactly like an ordinary miss, so an operator who sets, say, 512 MB on a 64-core box gets a cache that appears enabled and does nothing for most real blocks, with `debug_getCacheStatus` showing only a low entry count and a low hit rate.

**Suggested fix:** set `shards` explicitly so behaviour is host-independent, and/or log the effective per-shard budget once at startup plus a counter for admission-rejected inserts (compare `len()`/`weight()` across the insert, or track it in the lifecycle).

bin/debug-trace-server/src/block_data_cache.rs
```diff
@@ -0,0 +1,277 @@
...
+        let cache = BlockDataCache::new(one * 3 / 2);
+
+        for i in 1..=3u8 {
+            cache.insert(B256::from([i; 32]), Arc::clone(&data));
+        }
+        // quick_cache admits entries only within the byte budget (over-budget entries are
+        // rejected or evict older ones), so the cache never exceeds it.
+        assert!(cache.weight() <= one * 3 / 2);
+        assert!(cache.len() <= 2);
+    }
```
vincent-k2026, line 380: **Non-blocking — this test is vacuous: it passes on a cache that stores nothing.**

Measured on `18a127a`: one fixture entry weighs 95,589 B, this budget is 143,383 B, and the actual outcome is **`len = 0, weight = 0`** — nothing is ever admitted (see the shard thread on `:138`), so `weight() <= one * 3 / 2` and `len() <= 2` both hold trivially. With a 1 GB budget the same three inserts give `len = 3`.

As written it cannot distinguish "eviction respects the budget" from "the cache admitted nothing at all", which is exactly the failure mode the shard math produces.

**Suggested fix:** size the budget so entries are actually admissible (or set `shards` explicitly, per the other thread), then assert exact values — e.g. after three inserts with room for two, `len() == 2` and `weight() == 2 * one`. A second case pinning "entry larger than a shard's budget is rejected, and that rejection is counted" would cover the other half.

bin/debug-trace-server/src/block_data_cache.rs
```diff
@@ -0,0 +1,277 @@
...
+            .sum();
+
+        let weight = BlockDataWeighter.weight(&B256::ZERO, &Arc::new(data));
+        assert!(weight >= witness_bytes + contract_bytes + BLOCK_FIXED_OVERHEAD);
+
+        let mut heavier = fixture_block_data();
+        heavier.contracts.insert(B256::from([9u8; 32]), Bytecode::new_raw([0u8; 64].into()));
+        let heavier_weight = BlockDataWeighter.weight(&B256::ZERO, &Arc::new(heavier));
+        assert!(heavier_weight > weight, "adding a contract must increase the weight");
+    }
```
vincent-k2026, line 325: **Non-blocking — the access-list / authorization-list terms added in the last commit have no coverage.**

Measured on the fixture this test uses: 29 transactions, **0 access-list items and 0 authorization-list items**. So `tx.inner.access_list()` and `tx.inner.authorization_list()` always take the `map_or(0, ..)` branch here, and the assertions (`weight >= witness + contracts + BLOCK_FIXED_OVERHEAD`, and "adding a contract increases the weight") can't observe either new term.

The commit message says "weigh access/authorization lists", but nothing pins that behaviour or protects it from regressing.

**Suggested fix:** build a synthetic transaction carrying an access list with a known number of storage keys plus an EIP-7702 authorization, and assert the weight delta equals the expected `ACCESS_LIST_ITEM_BYTES` / `STORAGE_KEY_BYTES` / `AUTHORIZATION_BYTES` arithmetic — i.e. a test that fails if either term is dropped.

bin/debug-trace-server/src/block_data_cache.rs
```diff
@@ -0,0 +1,277 @@
...
+const AUTHORIZATION_BYTES: u64 = 192;
+
+impl Weighter<B256, Arc<BlockData>> for BlockDataWeighter {
+    fn weight(&self, _key: &B256, val: &Arc<BlockData>) -> u64 {
+        let witness = light_witness_memory_bytes(&val.witness) as u64;
+        let contracts: u64 = val
+            .contracts
+            .values()
+            .map(|code| ENTRY_OVERHEAD + code.bytes_slice().len() as u64)
+            .sum();
```
vincent-k2026: **Non-blocking — worth documenting: this term is charged against two independent budgets.**

The `Bytecode` values in `BlockData.contracts` are the same refcounted allocations held by `ContractCache`, so the same bytes are charged here *and* there. Evicting a `BlockData` entry while `ContractCache` still holds the code frees zero bytes. The direction is safe (the cache retains less real data than its budget suggests), and the type-level doc comment above acknowledges the over-count — but it means the `--block-data-cache-max-size` knob does not mean "this much block data", and that isn't stated in the README.

Combined with the witness term, which under-charges (BTreeMap / FxHashMap node overhead is excluded, as documented), the two errors point in opposite directions and neither is measured against real RSS. Given this is the mechanism bounding a 1 GB default, one recorded comparison of `weight()` against observed RSS on a real workload would be worth more than tightening the constants.

Minor, same function: the `other => other.len() as u64 * TX_OVERHEAD` arm at `:98` is unreachable — tracing requires `BlockTransactions::Full`, and a non-full block would fail long before it got here.

crates/stateless-common/src/witness_size.rs
```diff
@@ -76,6 +76,15 @@ impl WitnessSizeBreakdown {
     }
 }
 
+/// Approximate in-memory footprint of a [`LightWitness`] alone (KV entries + level map).
+///
+/// Unlike [`WitnessSizeBreakdown`], this needs no `MptWitness` and carries no fixed IPA
+/// term: a light witness holds no proof in memory. Suitable for byte-weighing caches that
+/// retain the witness without its MPT counterpart.
+pub fn light_witness_memory_bytes(light: &LightWitness) -> usize {
+    light.kvs.len() * SALT_KV_BYTES + light.levels.len() * SALT_LEVEL_BYTES
```
vincent-k2026, line 87: **Non-blocking — this PR promotes an observability-grade estimate into the enforcer of a memory bound; the constant should be derived rather than hand-maintained.**

`SALT_KV_BYTES` is 103, documented as "`SaltKey` (8 bytes) plus `Option<SaltValue>` (~95 bytes)". Measured: `size_of::<SaltKey>() = 8`, `size_of::<Option<SaltValue>>() = 95`, and `size_of::<(SaltKey, Option<SaltValue>)>() = 104` after padding. The 1-byte gap is noise (0.97%).

What matters is that this is a hand-copied approximation of a type layout in an external crate (`salt`), and until this PR it only fed metrics. Now it decides how much memory a 1 GB cache actually retains, so a `salt` bump that adds a field to `SaltValue` would silently make the cache over-retain with no signal anywhere.

**Suggested fix:** `const SALT_KV_BYTES: usize = size_of::<(SaltKey, Option<SaltValue>)>();` so the constant tracks the type, and keep the comment for the map-node overhead that is deliberately excluded.

bin/debug-trace-server/src/data_provider.rs
```diff
@@ -304,20 +312,36 @@ impl DataProvider {
...
+    }
+
+    /// Drops a block from the memory cache after its data failed execution: a decodable but
+    /// incomplete witness would otherwise stay pinned, failing every retry until eviction —
+    /// dropping it lets the next request refetch from the endpoints.
+    pub fn evict_block_data(&self, block_hash: &B256) {
+        if let Some(cache) = &self.block_data_cache {
+            cache.remove(block_hash);
+        }
+    }
```
vincent-k2026: **Non-blocking — eviction has no metric.**

Given the client-triggerable eviction path described on `rpc_service.rs:491`, an operator currently has no way to see that it's happening: no counter, nothing in `debug_getCacheStatus`. A counter labelled by method (and, once the discrimination rule lands, by reason) would make both the abuse case and the "our upstream is serving bad witnesses" case visible.

bin/debug-trace-server/src/main.rs
```diff
@@ -448,9 +461,18 @@ async fn main() -> Result<()> {
...
     let contract_cache = Arc::new(ContractCache::new(contract_store));
 
+    let block_data_cache = if args.block_data_cache_max_size == 0 {
+        warn!("Block-data cache disabled (max size = 0); every request re-resolves block data");
+        None
+    } else {
+        debug!(max_bytes = args.block_data_cache_max_size, "Block-data cache initialized");
+        Some(Arc::new(BlockDataCache::new(args.block_data_cache_max_size)))
+    };
+
```
vincent-k2026, line 512: **Non-blocking — two knob issues.**

1. **No lower bound.** Because of the per-shard admission threshold (see `block_data_cache.rs:138`), a value that is merely *small* rather than zero produces a cache that reports itself enabled and admits nothing. `warn!` only fires on exactly `0`. Rejecting implausibly small values at parse time — or at least logging the computed per-shard budget and the resulting maximum cacheable entry size here — would turn a silent misconfiguration into a startup-time signal.
2. **`BLOCK_DATA_CACHE_ESTIMATED_ITEMS` is hard-coded at 1024** while `max_bytes` is configurable; the response cache exposes its `estimated_items` as a flag. At 8 GB the internal sizing still assumes ~1024 entries. Either scale it from `max_bytes` and a typical entry weight, or expose it for symmetry with the response cache.

(Also the same `0`-means-disabled magic value as the response cache — if the explicit-disable-flag suggestion from #161 is taken, both should move together.)

README.md
```diff
@@ -102,6 +102,11 @@ Two operating modes:
...
 JS tracers, `muxTracer`, and struct-logger requests with non-default flags bypass the cache and are recomputed on every request.
 
+**Block-data cache:**
+A bounded in-memory `BlockData` cache keyed by block hash sits between the response cache and the local DB / RPC tiers, so requests that miss the response cache — other tracer variants of a block, and especially `debug_traceTransaction` calls for different transactions of the same block, which are never response-cached — reuse one witness fetch and contract resolution.
+Block-number lookups still resolve number → hash against the DB or upstream first, so canonicality is never cached and reorgs need no invalidation here.
+Size it with `--block-data-cache-max-size` (default `1GB`; `0` disables); blocks heavier than a cache shard's budget are never admitted, so very large blocks bypass it instead of thrashing it.
```
vincent-k2026: **Non-blocking — two things a reader can't currently get from this section.**

1. "blocks heavier than a cache shard's budget are never admitted" is accurate but not actionable: the shard budget is `max_bytes / (available_parallelism() * 4)`, reduced further by the hot/cold split. Measured, that's ~31 MB per entry at the 1 GB default on a 12-core host and ~6 MB on a 64-core host. Please state the resulting maximum cacheable entry size (or pin the shard count so the number is stable), since it drops as the host gets bigger.
2. The combined footprint isn't stated anywhere: response cache 1 GB + block-data cache 1 GB + contract cache 512 MB is **~2.5 GB by default**, on top of ordinary heap. The PR description says "+1GB RSS", which is true in isolation but not the number an operator sizing a box needs. Consider a 512 MB default here, or a one-time startup warning when the three budgets together exceed some fraction of system memory.

vincent-k2026 (changes requested):
## Review

**In one line:** the skeleton is right — keying by block hash, never caching canonicality, inserting exactly once inside the single-flight future all check out. The problems are concentrated in two places: the "evict on execution failure" hook added in the last commit catches none of the cases it was written for while letting any client drop other people's cache entries, and the cache advertises a byte bound that can't be computed from the flag, isn't covered by its test, and depends on the host's CPU count.

### What's genuinely good

1. **Keying by hash is the right answer.** An entry is an immutable fact about that hash, by-number lookups resolve number→hash first, and therefore reorgs need no invalidation hook at all — an orphaned hash's entry simply becomes unreachable and ages out. This is exactly the discipline the response cache in #161 should adopt; I've said so over there.
2. **The "inserts exactly once even if the primary caller is cancelled" claim holds.** I walked it: the insert sits inside the shared future after a successful fetch and is guarded by `?` (`data_provider.rs:641`), so failures never insert; a coalesced waiter still drives it to completion if the primary is cancelled; if *everyone* cancels, the future is dropped and `InFlightGuard` unconditionally clears the in-flight entry, leaving nothing poisoned.
3. **Three of the new tests actually discriminate**: DB hit populating memory and the second read skipping the DB (with `Arc::ptr_eq` proving one allocation), a memory hit short-circuiting ahead of a hanging upstream, and the disabled-cache path re-reading the DB. Those fail if the feature breaks.

### Blocking

- `rpc_service.rs:491` — the eviction hook's discrimination signal is wrong in both directions.

### The "bounded" claim has three holes

Details are in the inline threads; the summary:

1. **The admission threshold is a function of the host's core count** (`block_data_cache.rs:138`). quick_cache shards, an entry must fit in a *single shard's* budget, and the default shard count is `available_parallelism() * 4`. Measured on a 12-core box: one entry needs `max_bytes >= 33x its own weight` to be admitted at all. At the 1 GB default that's ~31 MB max cacheable block on 12 cores, ~6 MB on 64 cores, ~4 MB on 96 cores — **more cores means a smaller admissible block**, and a trace server runs on big machines. Rejected inserts are indistinguishable from ordinary misses.
2. **The per-KV weight constant is a hand-copied approximation of an external crate's type layout** (`witness_size.rs:85`). `SALT_KV_BYTES` is 103; `size_of::<(SaltKey, Option<SaltValue>)>()` measures 104. The 1-byte gap is noise — the point is that this PR promotes an observability-grade estimate into the enforcer of a memory bound, so a `salt` bump can silently make the cache over-retain with no signal.
3. **Contract bytecode is charged twice** (`block_data_cache.rs:72`). The `Bytecode` in `BlockData` and the one in `ContractCache` are the same refcounted allocation, so evicting a `BlockData` entry can free zero bytes. Conservative for RSS, but "1 GB" does not mean "1 GB of block data".

### Why the tests didn't catch any of this

- `eviction_respects_byte_budget` is vacuous — measured `len=0, weight=0` (see the inline thread).
- The access-list / authorization-list weight terms added in the last commit have zero coverage — the fixture block has 29 transactions with 0 access-list and 0 authorization-list items, so both new branches always take the `map_or(0, ..)` side.
- None of the five `evict_block_data` call sites has handler-level coverage; nothing asserts *when* eviction should and should not happen, which is why both directions of the bug survived.

### Operational

- Default memory across the three caches is **~2.5 GB** (response 1 GB + block-data 1 GB + contract 512 MB — `ContractCache::new` uses `ContractCacheConfig::default()`). The description says "+1GB RSS"; the combined footprint isn't documented anywhere.
- `evict_block_data` has no metric, so the amplification vector above would be invisible.
- `--block-data-cache-max-size` has no lower bound; combined with hole 1, a too-small value silently disables the cache while `warn!` only fires on exactly `0`.
- `BLOCK_DATA_CACHE_ESTIMATED_ITEMS` is hard-coded at 1024 while `max_bytes` is configurable — inconsistent with the response cache, which exposes `estimated_items` as a flag.

### Pre-existing, just noting

`get_block_data:378-382` uses `if let Ok(Some(hash)) = db.get_block_hash(..)`, silently dropping a DB `Err` and falling through to upstream, while its sibling `get_block_data_from_db:559-568` warns explicitly and documents why. Untouched by this PR, so not a request here.

### Things I checked that are fine

Listing these so it's clear they were looked at:

- `get_block_data_for_tx` does reach the memory tier (`:495`), so the PR's headline benefit holds. It still pays one upstream `eth_getTransactionByHash` per request, so what's saved is the witness fetch plus contract resolution, not the whole round trip.
- `light_witness_memory_bytes` is **not** missing a parent-commitments term — a `LightWitness` never materializes them (`witness_size.rs:57-60`), so kvs + levels is the correct set.
- I suspected the number→hash resolution in `get_block_data` duplicated the one in `is_canonical` and should be a shared memoized resolver. It shouldn't: the two treat "DB has no hash for this number" in opposite ways (fall through to upstream vs. treat as canonical below the retained range) and use different deadlines. Merging them would break the canonicality guard.

### Suggested landing order

1. **Required before merge:** the discrimination rule at `rpc_service.rs:491`.
2. Cheap and worth doing together: shard/rejection observability, the `size_of`-derived constant, an eviction counter.
3. Tests: make the budget test discriminate, cover the two new weight terms, and add handler-level tests for the eviction rule (that last one is really the acceptance criterion for item 1).
4. Docs: combined memory footprint, and what "a cache shard's budget" actually resolves to.

### How I verified

`cargo nextest run -p debug-trace-server` @ `18a127a` in a separate target dir: 96 tests, all green. The numbers above come from three instrumented probes in a scratch worktree (entry weight 95,589 B; binary search for the admission threshold; `size_of` of the salt KV pair; access-list/authorization counts in the fixture), reverted afterwards. I also read every per-transaction failure path in `tracing_executor.rs`, confirmed `ContractCacheConfig::default()` is 512 MB, and checked quick_cache 0.6.16's `Cache::with` / `OptionsBuilder::build` to confirm the shard default is never overridden here.

Note on stacking: base is still #161's branch, and #161 has open changes requested — worth settling that one's direction before retargeting.

Happy to be corrected anywhere, especially if the eviction hook was aimed at a failure mode I haven't reconstructed.

README.md
```diff
@@ -118,6 +118,13 @@ The memo holds `--canonical-hash-memo-capacity` entries (default 8M, roughly 80
...
+Size it with `--block-data-cache-max-size` (default `1GB`; `0` disables); the cache uses a fixed 4 shards, so the largest cacheable entry is `max_bytes / 4` on every host (~256 MB at the default), and an insert the cache does not retain is counted (`cache_admission_rejects_total`) and logged rather than masquerading as a miss.
+The byte budget is an estimate of retained payload, not exact RSS: map/allocator overhead is under-counted, and contract bytecode is charged here *and* in the contract cache (the same refcounted allocations), so evicting an entry can free less than its accounted weight.
+Combined default memory across the three caches is roughly 2.5 GB — response cache 1 GB + block-data cache 1 GB + contract cache 512 MB — on top of ordinary heap; size hosts (or lower the knobs) accordingly.
```
vincent-k2026: **Non-blocking — this predates the #161 merge and now omits a fourth cache.**

Since `liquan/fix/response-cache-idempotency` was merged in at `ebf2200`, the binary also carries the canonical-hash memo: `DEFAULT_CANONICAL_HASH_MEMO_CAPACITY = 8_000_000`, which the #161 description prices at ~0.6 GB if it ever fills. So "the three caches ... roughly 2.5 GB" is now short by one entry.

Suggested:

> Combined default memory across the four caches is roughly 3.1 GB — response cache 1 GB + block-data cache 1 GB + contract cache 512 MB + canonical-hash memo ~0.6 GB (8M entries, filled lazily; the measured working set is 50–130k heights, ~10 MB) — on top of ordinary heap.

Worth keeping the "filled lazily / measured working set" clause, otherwise 0.6 GB reads scarier than it is. Classic post-merge doc drift rather than anyone's oversight.

bin/debug-trace-server/src/block_data_cache.rs
```diff
@@ -0,0 +1,376 @@
...
+/// Shard count, pinned so the per-shard byte budget — the admission ceiling for a single
+/// entry — is `max_bytes / 4` on every host. quick_cache's default shard count scales
+/// with the core count, which would shrink the largest admissible block as machines get
+/// bigger; four shards keep contention negligible at trace-server request rates while
+/// leaving the ceiling far above any realistic block.
+pub(crate) const BLOCK_DATA_CACHE_SHARDS: usize = 4;
+
+/// Assumed typical in-memory size of one resolved block, used only to derive the
+/// estimated-items capacity hint from the byte budget.
+const EXPECTED_BLOCK_DATA_BYTES: u64 = 4 * 1024 * 1024;
```
vincent-k2026: **Non-blocking — where does 4 MB come from?**

This derives the estimated-items hint (`1 GB / 4 MB = 256`), which quick_cache uses for its initial sizing and, via `ghost_allocation`, for how many recently-evicted keys it tracks for admission decisions. If real `BlockData` weights are much smaller, the hint is off by an order of magnitude and S3-FIFO's admission quality degrades (correctness is unaffected).

What makes it stand out is that every other constant across these two PRs is backed by measurement — the 79% / 4–5% request split, the 93.7% repeat rate, 84–135 B/row, the 2.3–2.5 GiB full-index figure. This one looks estimated. For calibration: the synthetic fixture in these tests weighs ~95 KB, i.e. ~40x below the assumption.

You already have the 3.6-day replay; a p50/p90 of the real `block_data_weight` would either confirm 4 MB or replace it, and a one-line "measured at ..." comment would put it on the same footing as the rest.

bin/debug-trace-server/src/block_data_cache.rs
```diff
@@ -0,0 +1,376 @@
...
+            self.admission_rejects.fetch_add(1, Ordering::Relaxed);
+            self.metrics.record_admission_reject();
+            debug!(block_hash = %hash, weight, "Block-data cache did not retain insert");
+        }
+        self.refresh_size_gauges();
+    }
+
+    /// Removes an entry, e.g. when its witness turned out to be unusable at execution time —
+    /// a decodable but wrong upstream response must not stay pinned until eviction. Returns
+    /// whether an entry was present.
```
vincent-k2026, line 203: Nit — this can miscount. Between the `insert` and the `peek`, another thread's insert can legitimately evict this entry under pressure, and it gets recorded as an admission reject. The metric is a signal rather than an invariant so the noise is harmless, but a clause on the counter's doc ("a small number of false positives come from concurrent eviction between the insert and the check") would keep anyone from reading a non-zero value as proof of oversized entries.

vincent-k2026 (approved):
## Approving — verified against `0ab67b1`

Everything from my round-1 review is addressed, and the central item is solved more cleanly than what I asked for.

**The eviction rule now keys off a typed discriminant, and it's right in both directions.** `TraceError::{Request, Data}` plus `matches!(error, TraceError::Data(_))` at the call site means:

- The case the hook was written for is now actually caught: every tracer's `run_transaction` failure returns `Err(BlockReplayFailed)` instead of degrading into an `Ok` full of `TraceResult::Error` entries. The justification in the comment holds up — a canonical block's transactions all execute (reverts included) against correct state, so a replay failure really does mean the data is bad.
- The case that shouldn't have triggered it no longer does: mux/JS config parse and inspector construction map to `TraceError::Request`, so a client can't drop hot entries with a cheap invalid request.
- `trace_failure_evicts_block_data_only_for_data_errors` pins the rule at the handler layer, in both directions — the coverage whose absence let both directions of the bug survive.
- The client-visible consequence (`debug_traceBlock*` on unusable data now returns `-32000` rather than `200` with per-tx error entries) is called out in the description with its rationale. Agreed it's the right trade: the old shape was a partial wrong answer dressed as success.

**A false alarm I chased and am retracting**, so nobody "fixes" something that is already correct: I suspected mux and JS still degraded per-tx execution failures, because three `TraceResult::Error` sites remain in their branches. They're at `try_into_mux_frame`, the per-tx `JsInspector::with_transaction_context`, and `json_result` — tracer-output and per-tx construction failures, none of them data-attributable. Both tracers' `run_transaction` arms do abort (`tracing_executor.rs:603` and `:695`). The rule is applied consistently; no change wanted.

### The rest, checked rather than assumed

- **Shard count pinned to 4** with the reasoning written down, so the admission ceiling is `max_bytes / 4` (~256 MB) on every host instead of shrinking as machines get bigger. Exactly the fix.
- **Rejected inserts are now visible**: `peek` after insert → counter + metric + debug log, and `oversized_entry_is_rejected_and_counted` pins the path.
- **The budget test discriminates now.** Single shard so the arithmetic is deterministic, and it asserts exact values (`entry_count == 2`, `total_bytes == one * 2`) instead of inequalities that a cache admitting nothing would also satisfy.
- **The access-list / authorization terms have real coverage**: a synthetic transaction with 3 storage keys and 2 authorizations, asserting the *exact* weight delta against the constants — it fails if either term is dropped.
- `SALT_KV_BYTES` is now `size_of::<(SaltKey, Option<SaltValue>)>()`, so the bound tracks the type instead of a hand-copied number.
- `block_data_evictions_total` per method; `estimated_items` derived from the byte budget rather than hard-coded; README documents the weigher's under-count and the bytecode double-charge.
- The `StaticHashStore` → `StubBlockStore` refactor with read counters is a good call — "the memory-served read performs zero tip reads" is a much better assertion than "it returned something".
- `cargo nextest run -p debug-trace-server -p stateless-db` @ `0ab67b1`: **116 passed, 0 failed**.

### Non-blocking follow-ups

1. `README.md:126` — the combined-memory figure predates the #161 merge and now omits a fourth cache. Details inline.
2. `block_data_cache.rs:49` — where does `EXPECTED_BLOCK_DATA_BYTES` come from? Details inline.
3. `block_data_cache.rs:194` — the `peek`-based reject detection can miscount a concurrent legitimate eviction; worth a clause in the comment. Details inline.

None of these gate the merge. Good round.

bin/debug-trace-server/src/block_data_cache.rs
flyq: Fair catch — 4 MB was estimated, not measured. Measured now (563b397): `block_data_weight` over the 18 mainnet fixture blocks, with a production-shaped contracts map (only witness-referenced codes), gives ~70–360 KB for quiet blocks (~25 txs) and ~0.5–1.0 MB for the load-test-era ones (~600 txs) — your ~40× suspicion was right in direction. The constant is now 256 KiB (near the sample median) with the measurement in the comment, so the estimated-items hint at the 1 GB default goes 256 → 4096, matching the ~4k entries the budget actually holds. A p50/p90 over the 3.6-day replay would be better data still, but that dataset records request shapes rather than resolved `BlockData` weights, so the fixture blocks are what's measurable today.

README.md
flyq: Updated in 563b397 with your suggested framing: four caches, ~3.1 GB combined default — memo priced at its 8M-entry cap, with the fills-lazily / measured 50–130k-height (~10 MB) working-set clause kept so the 0.6 GB reads as the bound it is.

bin/debug-trace-server/src/block_data_cache.rs
flyq: Added in 563b397 — both the `insert()` doc and the counter's field doc now say a concurrent eviction racing the retention check can count as a reject, i.e. the counter is a pressure signal, not an exact tally of oversized entries.

bin/debug-trace-server/src/rpc_service.rs
flyq: One post-approval change you should know about, since your retraction explicitly said "no change wanted" there: `19f1119` (a cleanup pass that landed after the commit you reviewed) reclassified the two `try_into_mux_frame` failure sites from Request / per-tx `TraceResult::Error` to **Data** (abort + evict).

Reasoning: `try_into_mux_frame`'s only error channel is `DB::Error` — it runs the prestate collection against the witness DB internally — i.e. the same failure source as the plain prestate path, which the code you approved already classifies as Data. Leaving mux-frame as Request kept the eviction defeated for a witness gap surfacing through a mux prestate sub-tracer (the "eviction misclassification" the metric doc warns about). The per-tx `JsInspector` re-creation and `json_result` sites remain per-tx Request-level entries exactly as you described — JS output errors aren't cleanly attributable, and JS never enters the response cache. If you disagree with the mux-frame call, reverting to per-tx entries there is a two-line change.

Also since `0ab67b1`, the same pass moved classification into named constructors (`request_error` / `frame_build_error` / `tx_replay_failure`) with `trace_failure_to_rpc_err` as the sole metric+eviction+render conversion, so a new site picks a constructor by failure kind instead of hand-assembling a variant. Your three follow-ups are addressed in `563b397`.

vincent-k2026 (approved):
Approving the current head (`22fa5ee`). `cargo test --workspace` green here (99 bin + 39 common + 17 db). Re-reviewed the post-approval delta, and in particular the `try_into_mux_frame` call you flagged.

**On the mux-frame reclassification (`19f1119`): your call is correct, keep it.** I went to the type rather than the reasoning. `revm-inspectors 0.27.3`, `src/tracing/mux.rs:102-107`:

```rust
pub fn try_into_mux_frame<DB: DatabaseRef>(
    &self, result: &ResultAndState<impl HaltReasonTr>, db: &DB, tx_info: TransactionInfo,
) -> Result<MuxFrame, DB::Error>
```

`DB::Error` is the only channel, and in the body (`:108-158`) the single `?` is `geth_prestate_traces` at `:126` — the Call / FlatCall / Noop / FourByte arms all build their frame from an already-constructed inspector with no fallible step. So there is no "sub-tracer built fine but produced an unrepresentable frame" path to worry about: representability isn't checked here, and every request-attributable mux failure is already caught upstream at construction (`into_mux_config` → Request `tracing_executor.rs:244`, `MuxInspector::try_from_config` → Request `:232`). By the time frame-build runs the config is valid and the stage only reads state, and that `DB::Error` roots in `WitnessDatabaseError` (`evm_database.rs:94`) — the same source as the plain prestate path you already approved as Data. Classifying it Data does not reopen the client-triggerable-eviction hole.

Keeping JS at Request is the right complement, and for the right reason: `json_result` (`:696`, `:881`) genuinely has two channels — a client JS throw *or* a JS-initiated `db` read — so it's the one ambiguous site, and Request is the safe direction for a shape that's client-controlled and never response-cached. Mux has no such second channel. The line is drawn in the right place.

**The one thing worth knowing: that reclassification is the only change in the PR with no discriminating test.** Both `trace_block_discriminates_data_and_request_errors` (`tracing_executor.rs:1112`) and `trace_failure_evicts_block_data_only_for_data_errors` (`rpc_service.rs:992`) use `tracer_config: {"bogusTracer": {}}` for their Request case, which fails at `into_mux_config` (`:244`) and never reaches `try_into_mux_frame`. Flip `:630`/`:853` back to `request_error` and both tests still pass. Since the type argument above settles correctness I'm not blocking on it, and I appreciate that a real fixture is awkward here (you'd need a witness that replays execution but fails the post-execution prestate read). Noting it so the gap is on the record rather than discovered later.

Everything else I checked holds up. Eviction consistency across entry points is genuinely fixed — all five call sites route through `trace_failure_to_rpc_err` (`rpc_service.rs:198`, single `evict_block_data` caller), the old "debug block paths don't evict but parity does" split is gone, and a bad witness now surfaces as `TraceError::Data` on every tracer including JS via `tx_replay_failure` (`:721`). Response-cache poisoning is closed by ordering: `compute_*` propagates before `insert_cache` (`:521`→`:523`, `:569`→`:573`, `:693`→`:705`), and mux/JS carry `variant = None` so they're structurally uncacheable anyway.

On the weigher: no unbounded uncharged field exists. What's uncharged is `extra_data`, `uncles`, `withdrawals` and blob versioned hashes — all consensus-bounded and negligible. The worst-case ratio comes from a half-full `kvs` BTreeMap (~2x), the `levels` term (charged 5 B/entry vs ~10 B real for a hashbrown `(u32,u8)` slot), and analyzed-bytecode jump tables (~13%), so real RSS stays within a small constant factor of the reported weight — the right shape for a bound. `SALT_KV_BYTES` is genuinely `size_of`-derived (=104, `SaltValue` is an inline `[u8;94]` with no hidden heap). The insert/guard ordering in single-flight is leak-safe on cancel and error paths, a pure memory hit really opens zero redb transactions, and the four-cache 3.1 GB figure matches the code defaults exactly.

Three non-blocking follow-ups:

**Eviction amplification is real but bounded.** On a deterministically-bad-but-decodable witness, a hot block goes from ~0 upstream calls per request (cached, fast in-executor fail) to ~4 round-trips per request (evict → full `do_fetch_block_data`). Three things keep it from escalating: single-flight caps the concurrent blast to one fetch per hash, a fetch *timeout* surfaces as `DataProviderError` before the executor and so never triggers evict/refetch churn, and the amplification is a constant factor rather than compounding. Your deferred negative-caching-with-backoff is the right eventual fix; suppressing eviction for entries fetched within the last N ms would also do it.

**`--block-data-cache-max-size` still has no lower bound, and the per-shard budget is only `debug!`-level** (`main.rs:509`). A small-but-nonzero value yields a cache that reports itself enabled and admits nothing; it's observable now via `cache_admission_rejects_total`, but not at a log level anyone runs in production. Consider `warn!` when `per_shard_budget` falls below a sane block floor.

**The `levels` term is ~2x under-charged** (`witness_size.rs:87`). Bounded by buckets-touched and it folds into the documented "map overhead excluded" caveat, but it's the second under-charge stacked on the BTreeMap one, so either widen the constant or state the 2x explicitly.

Nit: `MuxInspector::try_from_config(...).unwrap()` on the block path (`tracing_executor.rs:616`) vs the fallible `mux_inspector(...)?` on the tx path (`:835`) — both safe since the config already parsed once, but prefer `?` in both.

Troublor (approved):

PR #167, opened 2026-08-01
perf(debug-trace-server): serve cache hits as raw JSON bytes

flyq:
## Summary
Handlers now return `RawJson` (`Arc<RawValue>`, `response_cache.rs:299`) instead of `serde_json::Value`: the trace output is serialized exactly once (`to_raw_value`, `rpc_service.rs:395`), the response-size metric reads the byte length, the response cache shares the same bytes by `Arc`, and jsonrpsee splices them verbatim into the reply. A cache hit becomes an `Arc` clone — zero parsing, zero copies, zero re-serialization. Two silent-failure paths are fixed along the way: a serialization failure used to store an **empty string** in the cache (`to_string(..).unwrap_or_default()`), and a hit-side parse failure returned **null** (`unwrap_or(Value::Null)`) — both now propagate as errors through `RawJson::try_new`.

## Root cause
The cache stored pre-serialized strings but its interface to jsonrpsee was `serde_json::Value`, so both sides of the cache paid full tree work: a hit re-parsed the stored string into a `Value` and jsonrpsee re-serialized it (measured ≈7 ms CPU per MB of response — a ~20 MB block-level callTracer hit burns ~140 ms server CPU, see the baseline below; 93.7% of production traffic is repeat-after-success, i.e. exactly this path); the fill path serialized the same response three times (`value.to_string().len()` for the size metric, `CachedResponse::new`, jsonrpsee).

## Measured baseline (v2.0.15 without this PR, 2026-08-01)
Bench on the tko staging deploy (response cache enabled), 51 load-test-era blocks 6907000–6907050 — 36.7k–44.1k tx each (median 41,779), callTracer responses avg 20.0 MB, 1020 MB per round:

| round | status | total p50 / p90 / max | server CPU p50 (`x-execution-time-ns`) |
| --- | --- | --- | --- |
| 1 (cold) | 51/51 OK | 1244 / 1527 / 1760 ms | 721 ms (incl. ~337 ms EVM replay) |
| 2–3 (cache hits) | 51/51 OK | 168–188 / 305–402 / 523–803 ms | **140 / 135 ms** |

Raw data: `validator-data/debug_trace_server/bench/new/monster51_files_tko2.tar.gz`, analysis in `bench/old-vs-new-report-2026-08-01.md` §7.

## Validated on tko (2026-08-03, combined with #174)
Same segment, same box, this PR deployed together with #174 (response compression); full data in `validator-data/.../bench/0803/`, analysis in `bench/old-vs-new-report-2026-08-01.md` §9:

| | baseline (§7) | with this PR |
| --- | --- | --- |
| warm-hit server CPU p50 | 135–140 ms | **12 ms (−91%)** |
| warm-hit total p50 | 188 ms | **125–133 ms** (remainder = 20 MB loopback transfer — hit path is now IO-bound as designed) |
| cold r1 p50 / CPU p50 | 1244 ms / 721 ms | **1047 ms / 568 ms** (fill-path 3→1 serializations) |

Zero regressions vs baseline (`--compare`: 0 regressed, 51/51 OK all rounds); cold-round witness fetches confirm a genuinely cold r1. Composes cleanly with #174: encoded responses keep the 12 ms hit CPU (compression runs outside the handler, pipelined with the write).

## Fix
- `RawJson` in `response_cache.rs`: `Serialize` splices the stored bytes verbatim (`:338`); the cache stores it directly as its value type (`get`/`insert` at `:483`/`:512`), deleting `CachedResponse`.
- `compute_block_trace` produces `RawJson` via `serde_json::value::to_raw_value` — one serialization, no validation re-scan of our own output; the size metric reads `byte_len()` (`rpc_service.rs:395`).
- Both tx handlers (`debug_traceTransaction`, `trace_transaction`) drop their `to_value` + `to_string` double serialization; parity not-found returns `RawJson::null()`.
- `insert_cache` keeps its slow-insert timing (restored in e0cbc21 after review): the insert is an `Arc` clone, but at the weight ceiling `quick_cache` evicts synchronously and each eviction takes the index lock, so an insert-side stall stays attributable.
- `debug_getCacheStatus` keeps `serde_json::Value` (tiny payload, not cached).

## Testing
- `raw_json_serializes_verbatim` (`response_cache.rs:985`) pins the splice property the design rests on.
- `raw_json_passes_through_jsonrpsee_verbatim` (`rpc_service.rs:914`) proves it end-to-end through a real `RpcModule` reply.
- `shares_bytes_with` assertions pin that hits serve the inserted bytes without copying (`rpc_service.rs:905`, cache insert/get test).
- Full workspace: 20/20 test targets green; fmt/clippy/cargo-sort clean.

## Notes
- Both wins are live in production configs: the fill-path win (3 serializations → 1) applies to every response, and prod now runs with the response cache enabled (32 GB on ore/fra; the old `ESTIMATED_ITEMS=0` disable convention is rejected at startup since v2.0.15), so the hit-path win applies immediately on deploy. Verify via `debug_trace_cpu_time` / hit p50 against the tables above.
- `serde_json` gains the `raw_value` feature in this bin only.
- Cached replies now return the exact original bytes instead of a re-serialization of them.
- Wire-visible byte change: response JSON keys now serialize in **struct declaration order** (the old `Value` path emitted alphabetical order via its `BTreeMap`, `preserve_order` being off). Semantically irrelevant for JSON-RPC and closer to geth/reth, which also serialize from structs — but golden-file/snapshot comparisons will show the difference.

### DIFF
--- a/bin/debug-trace-server/src/rpc_service.rs
+++ b/bin/debug-trace-server/src/rpc_service.rs
@@ -27,7 +27,7 @@ use crate::{
         METHOD_DEBUG_TRACE_TRANSACTION, METHOD_TRACE_BLOCK, METHOD_TRACE_TRANSACTION,
         ResponseSizeMetrics, RpcGlobalMetrics, SingleFlightMetrics,
     },
-    response_cache::{CachedResource, RequestShape, ResponseCache, ResponseVariant},
+    response_cache::{CachedResource, RawJson, RequestShape, ResponseCache, ResponseVariant},
     tracing_executor::TraceError,
 };
 
@@ -44,7 +44,7 @@ pub trait DebugTraceRpc {
         &self,
         block_number: BlockNumberOrTag,
         opts: Option<GethDebugTracingOptions>,
-    ) -> RpcResult<serde_json::Value>;
+    ) -> RpcResult<RawJson>;
 
     /// Trace block execution by block hash.
     #[method(name = "traceBlockByHash")]
@@ -52,7 +52,7 @@ pub trait DebugTraceRpc {
         &self,
         block_hash: B256,
         opts: Option<GethDebugTracingOptions>,
-    ) -> RpcResult<serde_json::Value>;
+    ) -> RpcResult<RawJson>;
 
     /// Trace a single transaction execution.
     #[method(name = "traceTransaction")]
@@ -60,7 +60,7 @@ pub trait DebugTraceRpc {
         &self,
         tx_hash: B256,
         opts: Option<GethDebugTracingOptions>,
-    ) -> RpcResult<serde_json::Value>;
+    ) -> RpcResult<RawJson>;
 
     /// Query current response cache status.
     #[method(name = "getCacheStatus")]
@@ -72,12 +72,12 @@ pub trait DebugTraceRpc {
 pub trait TraceRpc {
     /// Parity-style block tracing (flat call traces).
     #[method(name = "block")]
-    async fn trace_block(&self, block_number: BlockNumberOrTag) -> RpcResult<serde_json::Value>;
+    async fn trace_block(&self, block_number: BlockNumberOrTag) -> RpcResult<RawJson>;
 
     /// Parity-style transaction tracing.
     /// Returns null (not error) when transaction is not found, matching mega-reth behavior.
     #[method(name = "transaction")]
-    async fn trace_parity_transaction(&self, tx_hash: B256) -> RpcResult<serde_json::Value>;
+    async fn trace_parity_transaction(&self, tx_hash: B256) -> RpcResult<RawJson>;
 }
 
 // RPC Watch Dog
@@ -224,7 +224,7 @@ impl RpcContext {
     /// *before* the cache lookup, all on one request deadline; on a miss, fetch block data
     /// by the resolved hash on the remaining budget. Slow prelude stages are warned about
     /// here, where they are measured; trace/serialize timing lives in
-    /// [`compute_block_trace`] and cache-insert timing in [`insert_cache`].
+    /// [`compute_block_trace`].
     async fn lookup_block_by_number(
         &self,
         method: &'static str,
@@ -293,7 +293,7 @@ impl RpcContext {
         data: &BlockData,
         method: &'static str,
         opts: GethDebugTracingOptions,
-    ) -> Result<serde_json::Value, jsonrpsee::types::ErrorObjectOwned> {
+    ) -> Result<RawJson, jsonrpsee::types::ErrorObjectOwned> {
         compute_block_trace(data, method, || {
             crate::tracing_executor::trace_block(
                 &self.chain_spec,
@@ -310,7 +310,7 @@ impl RpcContext {
 /// Outcome of [`RpcContext::lookup_block_by_number`].
 enum BlockLookup {
     /// Served straight from the response cache.
-    Cached(serde_json::Value),
+    Cached(RawJson),
     /// Cache miss: block data fetched by the resolved canonical hash, ready to trace.
     Fetched(Arc<BlockData>),
 }
@@ -374,14 +374,16 @@ fn data_provider_error_to_rpc_error(e: &DataProviderError) -> jsonrpsee::types::
 
 // Trace Computation Helpers
 /// Runs a block-trace executor closure and wraps the scaffolding shared by every
-/// block-level handler: EVM timing + metrics, JSON serialization, the response-size
-/// metric, and the slow-stage warning. Returns the typed [`TraceError`] so the caller can
-/// key cache hygiene off its data-vs-request discriminant before rendering an RPC error.
+/// block-level handler: EVM timing + metrics, the single JSON serialization (straight
+/// from the trace output into [`RawJson`] bytes — the tree is never rebuilt as a
+/// `serde_json::Value`), the response-size metric, and the slow-stage warning. Returns
+/// the typed [`TraceError`] so the caller can key cache hygiene off its data-vs-request
+/// discriminant before rendering an RPC error.
 fn compute_block_trace<T: serde::Serialize>(
     data: &BlockData,
     method_name: &'static str,
     run: impl FnOnce() -> Result<T, TraceError>,
-) -> Result<serde_json::Value, TraceError> {
+) -> Result<RawJson, TraceError> {
     let start = Instant::now();
 
     let results = run()?;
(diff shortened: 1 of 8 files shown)


bin/debug-trace-server/src/rpc_service.rs
```diff
@@ -446,30 +450,33 @@ fn check_cache(
...
     variant: Option<ResponseVariant>,
-    method_name: &'static str,
-    result: &serde_json::Value,
+    result: &RawJson,
 ) {
     let (Some(cache), Some(variant)) = (cache, variant) else {
         return;
     };
-    let t = Instant::now();
     cache.insert(resource, block_hash, variant, result);
```
Troublor, line 468: The serialization is gone from this path, but it was never the only cost here: `ResponseCache::insert` still does `index.insert`, then `quick_cache::Cache::insert` — which at the weight ceiling evicts synchronously, each eviction running `on_evict -> HashIndex::remove_key` under the index lock (`response_cache.rs:351-357`) — and finally `update_size_metrics()`. So "there is no multi-MB copy to time" is true, but "nothing left to time" isn't.

With the warning removed, an insert-side stall is now invisible: `compute_block_trace`'s slow-stage warning stays silent (trace and serialize are both fast), and the only symptom is unattributed total latency in `record_request_completion`. Suggest keeping the timer and the `method_name` parameter:

```suggestion
    let t = Instant::now();
    cache.insert(resource, block_hash, variant, result);
    let insert_ms = t.elapsed().as_millis();
    if insert_ms >= SLOW_STAGE_THRESHOLD_MS {
        warn!(
            method = method_name,
            block_hash = %block_hash,
            cache_insert_ms = insert_ms as u64,
            "slow response-cache insert"
        );
    }
```

(This restores the `method_name: &'static str` parameter and its five call sites; the doc comment above should lose the "no multi-MB copy to time" sentence.)

flyq: Restored in e0cbc21 — took your suggestion as-is: the timer, the `slow response-cache insert` warning, and the `method_name` parameter are back with the original signature and all five call sites updated. The doc comment now states the accurate reason instead of the deleted one: the insert is an `Arc` clone but not free — at the weight ceiling `quick_cache` evicts synchronously and each eviction takes the index lock. That regime isn't hypothetical for us: the fra deployment ran its cache at 29GB/32GB with sub-minute eviction churn earlier this week, which is exactly where an unattributed insert-side stall would have been invisible. Good catch.

vincent-k2026 (commented):
Reviewed against HEAD `e0cbc21`. Solid change; three things below.

## 1. The insert-timing restore left one doc statement behind

`insert_cache` has its timer and its `slow response-cache insert` warning back as of `e0cbc21`, but the doc change that accompanied their removal is still in the diff — `rpc_service.rs`, `lookup_block_by_number`:

```diff
-    /// [`compute_block_trace`] and cache-insert timing in [`insert_cache`].
+    /// [`compute_block_trace`].
```

So the code times inserts and the doc says it doesn't. The next person chasing an insert-side stall reads this and looks in the wrong place. Please drop this hunk.

Side effect worth knowing: #168 keeps the original wording, so **#168's version is the correct one** — dropping this hunk also removes the conflict between the two PRs on that line.

## 2. The response body's JSON key order changes on the wire — not mentioned in the description

`serde_json` is configured without `preserve_order` (`Cargo.toml:98`), so `Value::Object` is a `BTreeMap`:

- old path `to_value(...).to_string()` → keys in **alphabetical** order
- new path `to_raw_value(&results)` → keys in **struct declaration** order

Semantically irrelevant for JSON-RPC, and the new order arguably matches geth/reth better since they also serialize from structs — so this is probably an improvement. But it is a user-visible byte change, and anything downstream doing golden-file or snapshot comparison will break on it. Worth a line in the PR body.

## 3. Two silent-failure paths are fixed here, and the description sells this purely as a perf change

The old `CachedResponse::new` was `serde_json::to_string(value).unwrap_or_default()` — a serialization failure stored an **empty string** in the cache — and `as_value()` was `from_str(..).unwrap_or(Value::Null)`, turning a parse failure into a **null result**. Both now propagate through `RawJson::try_new`. That is more valuable than the latency numbers and should be in the summary.

## What's good

- The `Arc<Box<RawValue>>` choice is justified in a comment rather than left to be re-derived: `Arc<RawValue>` has an unsized pointee, so constructing it memcpys the whole body — exactly what this PR exists to avoid.
- `raw_json_passes_through_jsonrpsee_verbatim` drives a real `RpcModule`, so the property the entire design rests on (jsonrpsee splices the bytes rather than re-parsing) is pinned end-to-end instead of assumed.
- Switching the cache assertions from `assert_eq!(value)` to `shares_bytes_with` (`Arc::ptr_eq`) tests the actual property — that a hit is zero-copy — rather than that two values happen to be equal.
- Keeping the insert timed is right: the insert is an `Arc` clone now, but at the weight ceiling `quick_cache` evicts synchronously and each eviction takes the index lock, so there is still a stall worth attributing.

flyq:
@vincent-k2026 All three addressed: the leftover doc hunk is dropped in 07031ae (restores main's wording, which also removes the conflict with #168 on that line); the description's summary now names the two silent-failure fixes (empty-string cache fill / null hit result), and the notes call out the wire-visible key-order change with the golden-file caveat. Thanks for the thorough what's-good list — the `Arc<Box<RawValue>>` sizing point especially.

vincent-k2026 (approved):
Re-reviewed at `07031ae`. All three items from my last pass are closed — approving.

1. **Doc contradiction resolved.** `lookup_block_by_number`'s doc is back to naming `insert_cache` alongside `compute_block_trace`, so the pointer and the restored timer agree again, and it now matches main / #168 on that line — the cross-PR conflict I flagged is gone with it. Verified the timer and the `slow response-cache insert` warn are both live in `insert_cache`.
2. **Key-order change is in the body** with the golden-file/snapshot caveat and the geth/reth-parity rationale.
3. **Both silent-failure fixes are named in the summary** (empty-string cache fill, `null` hit result → propagated through `RawJson::try_new`).

One stale line left, body-only and not blocking the merge: under **## Fix**, "`insert_cache` loses its slow-insert timing — the insert is now an `Arc` clone with nothing left to time" is what `e0cbc21` reverted. Worth a one-line edit so the squashed commit message doesn't ship a claim the code contradicts.

FYI `pr-review` is red here and on every other open PR in the repo — the org `claude-pr-review` composite action dies with `ENOENT: no such file or directory, symlink` before it reviews anything. Infra, unrelated to this change; everything else is green.

Troublor (approved):

Troublor (approved):

batatmosfer-byte:
👍👌

śr., 5 sie 2026, 04:46 użytkownik liquan.mega ***@***.***>
napisał:

PR #169, opened 2026-08-01
fix(debug-trace-server): reject malformed tracerConfig in the executor

flyq:
## Summary
The executor's six tracerConfig parses (call/prestate/flatCall on the block path `tracing_executor.rs:528` and the single-tx path `:787`) now fail with `TraceError::Request` on a type-malformed config instead of silently falling back to default settings (`unwrap_or_default`). Pure defense in depth — no user-visible behavior change through the RPC surface.

## Root cause
The `-32602` rejection of malformed configs lives at the RPC boundary (`RequestShape::classify` in `classify_and_gate`), so the executor parse sites are unreachable with bad configs today — but the two layers were coupled only by an "unreachable in practice" comment. Any future entry point that skips the boundary gate would have traced with default settings and returned HTTP 200: exactly the silent-degradation behavior the boundary fix (vincent's review on #161) was written to eliminate. Deferred back then because the parse sites' functions returned the shared crate's `ValidationError`; the bin-local `TraceError { Request, Data }` that has since landed removed that blocker.

## Fix
Six `unwrap_or_default()` → `.map_err(|e| request_error("invalid <tracer> tracerConfig", e))?`, matching the muxTracer parse right above them (`:244`); the coupling comments become a defense-in-depth note. The boundary `-32602` remains the user-facing rejection; executor-level rejection surfaces through the existing `TraceError::Request` mapping (`-32000`), and — like all request-attributable failures — never evicts cached block data, so a cheap invalid request still cannot drop hot entries.

## Testing
`malformed_tracer_config_is_rejected_not_defaulted` (`tracing_executor.rs:1167`): all three config-reading builtins, on both the block and single-tx paths, reject a type-malformed config with a tracerConfig-naming `Request` error, and a well-formed config on the same fixture inputs still traces. The RPC layer's own `-32602` stays pinned by the existing `invalid_tracer_config_maps_to_invalid_params`. Workspace fmt/clippy clean, all debug-trace-server unit tests green (107).

## Notes
Out of scope, unchanged from the deferred-findings plan: threading classify's already-parsed typed config into the executor to remove the double parse — a larger interface change not worth bundling into this backstop.

### DIFF
--- a/bin/debug-trace-server/src/tracing_executor.rs
+++ b/bin/debug-trace-server/src/tracing_executor.rs
@@ -241,9 +241,47 @@ fn mux_config_and_inspector(
     let config = tracer_config
         .clone()
         .into_mux_config()
-        .map_err(|_| TraceError::Request("Invalid mux tracer config".into()))?;
+        .map_err(|e| request_error("invalid muxTracer tracerConfig", e))?;
     let inspector = mux_inspector(config.clone())?;
     Ok((config, inspector))
+}
+
+/// Parses a config-reading builtin's `tracerConfig` into its [`TracerKind`] — the one
+/// home for the strict-parse contract shared by the block and transaction dispatchers.
+///
+/// Defense in depth: RPC handlers already reject malformed configs via
+/// `RequestShape::classify` (-32602) before the executor runs, so from the current entry
+/// points this rejection is unreachable — but a malformed config must never silently
+/// trace with default settings if a future entry point skips the boundary gate. The
+/// executor guarantees only that much: its rejection surfaces as a request-attributable
+/// [`TraceError`], not as the gate's `-32602` wire code.
+fn builtin_tracer_kind(
+    builtin: GethDebugBuiltInTracerType,
+    tracer_config: &GethDebugTracerConfig,
+) -> Result<TracerKind, TraceError> {
+    let config = tracer_config.clone();
+    match builtin {
+        GethDebugBuiltInTracerType::CallTracer => config
+            .into_call_config()
+            .map(TracerKind::Call)
+            .map_err(|e| request_error("invalid callTracer tracerConfig", e)),
+        GethDebugBuiltInTracerType::PreStateTracer => config
+            .into_pre_state_config()
+            .map(TracerKind::PreState)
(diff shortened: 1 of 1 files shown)


vincent-k2026 (commented):
The cleanest one in this stack. No blockers from me.

1. **Six `unwrap_or_default()` silent-defaults removed.** A type-malformed `tracerConfig` previously traced with default settings and returned a result that *looked* normal — the caller had no way to know its config was dropped on the floor. Turning that into `TraceError::Request` is the right direction.

2. **Three near-identical arms in the block and transaction dispatchers collapse into one**, with the strict-parse contract living in `builtin_tracer_kind`. The old shape was three parallel copies where missing one on a future edit would have gone unnoticed.

3. **The test actually discriminates.** `malformed_tracer_config_is_rejected_not_defaulted` calls `trace_block` / `trace_transaction` directly, bypassing the RPC layer's `-32602` gate, so it exercises the new code rather than the pre-existing guard. Put `unwrap_or_default()` back and it goes red. The well-formed-config positive control alongside it is what makes it a real regression test instead of a "does it error" assertion.

4. **The doc is honest about reachability** — it states plainly that this rejection is unreachable from the current entry points and is defense-in-depth, rather than dressing a hardening change up as a bug fix. That's the right way to present a behaviorally-inert change.

flyq:
@vincent-k2026 Nothing to action — appreciated the careful read, especially verifying that the test discriminates (goes red on a restored `unwrap_or_default()`) rather than just asserting "some error".

vincent-k2026 (approved):
Approving at `d109d93` — unchanged since my review and I had no blockers.

Restating the reason it clears the bar: removing the six `unwrap_or_default()` calls turns a silently-ignored `tracerConfig` into an explicit rejection, the tests call `trace_block` / `trace_transaction` directly so they bypass the RPC layer's `-32602` gate and genuinely discriminate (restoring `unwrap_or_default()` turns them red), and the doc is honest that the new rejection is unreachable from the current entry point — defence in depth, not dressed up as a bug fix.

`pr-review` is red on every open PR in the repo (org `claude-pr-review` action fails with `ENOENT ... symlink` before reviewing) — infra, not this PR.

Troublor (approved):

PR #170, opened 2026-08-01
refactor: apply simplify-review cleanups across core, common, and the binaries

flyq:
## Summary
Two `/simplify` review passes (4 review agents each: reuse, simplification, efficiency, altitude) over `stateless-core` and `bin/stateless-validator`, with the surviving findings applied — net −73 lines. No functional change on canonical-block paths; behavior-adjacent changes are called out below.

## stateless-core
- Deleted the no-op hardfork reorder loop in `ChainSpec::from_genesis`: `MegaethGenesisHardforks::into_vec()` already yields canonical activation order (its literal is now documented as the single source of that order), and the "unknown fork" append branch was unreachable; the unused `mega_mainnet_hardforks()` template is gone with it.
- Removed the never-used EIP-3155 `trace_writer` parameter threaded through `replay_block` / `verify_and_replay` / `validate_block` / `validate_block_deriving_updates` (no caller here or in mega-reth ever passed `Some`), which also deletes the `run_plain` closure + cfg dance. **mega-reth impact**: its call sites drop a trailing `None` at the next version bump.
- `WitnessExternalEnv::new` / `from_light_witness` now share one `from_metadata_kvs` body; dead `LightWitnessExecutor::kvs()` accessor and a doubled doc block on `create_evm_env` removed.
- Reuse: the blob base-fee fraction comes from revm's `BLOB_BASE_FEE_UPDATE_FRACTION_CANCUN` instead of a hand-copied consensus constant; `gas_used` reads `BlockExecutionResult::gas_used` instead of recomputing from the last receipt.
- Hot path: `WitnessDatabase` reads use salt's `find()` + in-place decode and stack-allocated key encodings (`pub(crate)` helpers `encode()` also delegates to) — two heap allocations per witnessed state read and one per `bucket_id_for_*` call removed; dropped a per-miss `SaltValue` clone in `LightWitness::metadata`.
- Pipeline: `chain_advancer` runs `pre_advance` + `advance_chain` under `spawn_blocking` so redb commits (and the trace server's multi-MB block-data writes) no longer pin an async runtime worker; panics still propagate unchanged via `resume_unwind`. Also removed the `in_flight_blocks` mirror set, the `metas` lockstep vector, and the dead `_reason` parameter.
- `ChainStore` no longer carries the `ContractStore` supertrait (nothing consumed contracts through it); the forced stubs on the pipeline mock and the trace server's `StubBlockStore` are deleted.

## stateless-validator + shared crates
- R2 mode no longer hands the data endpoints to `RpcClient` as placeholder witness endpoints: the constructor accepts an empty witness list, and a witness call with zero providers panics with a structural message instead of silently retrying `mega_getBlockWitness` against data endpoints. The trace server is unaffected (clap already enforces a witness source at startup).
- `BackoffPolicy::schedule()` / `BackoffSchedule` in stateless-common now own the jittered-doubling arithmetic previously duplicated between the RPC round-robin loop and the R2 GET loop; both consume it (`fastrand` dep dropped from the bin).
- `install_prometheus_exporter` hoisted into `stateless_common::metrics`; both binaries call it and the `metrics-exporter-prometheus` dep moved from the bins into common.
- `BlockMeta::from_header` added in core, replacing the three defaulting header→meta projections (validator + two in debug-trace-server); the strict anchor-init site in `app.rs` keeps its explicit missing-withdrawals-root error by design.
- `verify_block_integrity` now runs on the blocking pool and encodes each tx exactly once (keccak of that encoding checks the hash, the same bytes feed the transactions-root trie) — previously 2 clones + 2 encodes per tx inline on runtime threads.
- Validator cleanups: `on_remote_height` fn-pointer injection removed; `run_with_signals` takes `report_validation: bool` instead of an `Option<String>` it only ever `.is_some()`'d; phantom `eth_getTransactionByHash` metric series dropped; test-only `set_anchor_block` deleted (the roundtrip test now exercises `reset_to_anchor`); `workers.rs` renamed to `runner.rs` (docs updated); integration mock handlers deduped onto shared lookup helpers (also fixing a drifted bare `unwrap` on malformed params); `make_block_meta` test fixture unified.

## Testing
`cargo fmt` / `clippy --workspace --all-targets --all-features` (0 warnings) / `cargo sort` clean; full workspace suite 328 passed / 0 failed; no-std check and no-std test compile (`--no-default-features`) pass.

## Notes
- The empty-witness-list change relaxes a public `RpcClient` constructor check (startup rejection → loud panic at first misuse); the constructor test pins the new contract.
- Known follow-up (next PR): hoist the block-execution env assembly shared by `replay_block` and the trace server's `TracingEnv::new`, resolving their `BlockLimits` drift.

## Review response (2026-08-04)

Behavior-change inventory (from review, kept in one place deliberately):
1. `witness_apis` may be empty; a witness call with zero providers asserts — currently unreachable (the R2 branch is taken first, and there is no second `get_witness*` call site in the validator). Assert-vs-typed-error is ledgered as a small follow-up PR.
2. `verify_block_integrity` moved into `spawn_blocking` — its interaction with `round_robin_with_backoff`'s per-attempt timeout (a healthy provider serving a large block could be classified as stalled while the blocking task runs on) is acknowledged and ledgered as its own follow-up PR rather than patched inside this one.
3. `verify_block_integrity` rewritten: `trie_hash()` → explicit `encode_2718` + `keccak256` (definitionally equal — `Encodable2718::trie_hash` is `keccak256(encoded_2718())`), removing N clones and a second encode pass.
4. `gas_used` switched from `receipts.last().cumulative_gas_used()` to `execution_result.gas_used` — **definitionally equivalent**: mega-evm's `finish()` computes `gas_used` as exactly `receipts.last().cumulative_gas_used()`, so both expressions read the same source regardless of how system transactions are accounted. Pinned by a `debug_assert_eq!` at the derivation site, exercised by the mainnet single-block integration test.
5. `chain_advancer` takes `Arc<S>`/`Arc<H>` and commits via `spawn_blocking`; panics are preserved via `try_into_panic` → `resume_unwind`.
6. `evm_database`: `plain_value()` → `find()`, dropping a heap allocation per state read.

Hardfork ordering now has enforcement: `test_mega_hardforks_iterate_in_activation_order` asserts the relative order of all ten MegaETH forks end-to-end through `from_genesis` (order is consensus-relevant: timestamp fork selection walks `forks_iter()` in insertion order), and the `into_vec` literal's comment points at the test.

**Breaking at the next tag** — mega-reth consumes `stateless-core` pinned at v2.0.14; nothing breaks until the bump, which needs coordination:
- `mega_mainnet_hardforks()` and `LightWitnessExecutor::kvs()` removed.
- `BLOB_GASPRICE_UPDATE_FRACTION` replaced by revm's `BLOB_BASE_FEE_UPDATE_FRACTION_CANCUN` — same value (3338477), a consensus parameter, verified equal.
- `validate_block` / `validate_block_deriving_updates` / `replay_block` drop the `writer` parameter, deliberately removing EIP-3155 trace output (no caller here or in mega-reth ever passed `Some`); mega-reth call sites drop a trailing `None` on the bump.
- `ChainStore: ContractStore` supertrait dropped — downstream payoff: mega-reth's `MegaValidatorChainStore` deletes its intentionally-inert `ContractStore` impl on the bump.


vincent-k2026 (commented):
Taken individually I agree with almost every change in here. My ask is about the packaging: this is labelled a cleanup, and it carries at least six behavior changes plus five public-API removals across 30 files. That makes each one un-reviewable in isolation, un-bisectable, and un-revertable independently.

Item 2 below is the concrete cost. Moving `verify_block_integrity` into `spawn_blocking` is a good change on its own, but it also moved verification inside `round_robin_with_backoff`'s per-attempt timeout, so a healthy provider serving a large block can be classified as stalled, retried, and leave the blocking task running behind it. As a standalone three-line PR that consequence is visible on sight; at position 2 of 6 in a 30-file cleanup it is not.

Suggest splitting into a genuinely behavior-preserving cleanup plus a handful of small, individually-motivated changes. Concretely, here is what I count as behavior change:

## 1. Behavior changes shipped under a refactor label

1. **`witness_apis` may now be empty, and a witness call with zero providers `assert!`s** — a new panic path in a long-running binary. I checked the reachability: `ValidatorFetcher::fetch` takes the R2 branch when `r2_witness.is_some()`, and there is no second `rpc_client.get_witness*` call site in `bin/stateless-validator`, so it is currently unreachable. It is also strictly better than the old "hand it the data endpoints as a placeholder" behavior, which silently pointed witness calls at the wrong endpoints. Still worth being deliberate about `assert!` vs a typed error in a validator that is expected to stay up.
2. **`verify_block_integrity` moved to `spawn_blocking`** — the per-attempt-timeout interaction described above.
3. **`verify_block_integrity` rewritten** — clones removed, `trie_hash()` replaced by explicit `encode_2718` + `keccak256`. These are equivalent (`Encodable2718::trie_hash()` is defined as `keccak256(encoded_2718())`) and it removes N clones plus a second encode pass, so it is a good optimization — of a **security-critical verification function**, which deserves its own PR and its own line in a changelog.
4. **`gas_used` now comes from `execution_result.gas_used` instead of `receipts.last().cumulative_gas_used()`.** This is the one I would most like to see argued explicitly. It feeds the header check, and MegaETH carries system transactions in `extra_data`, so whether the two agree depends on whether the executor folds system-transaction gas into its total. If they can disagree, this is a consensus divergence. Please state why they are equivalent and add an assertion test pinning it.
5. **`chain_advancer` takes `Arc<S>`/`Arc<H>` and commits via `spawn_blocking`** — a concurrency model change. The panic handling (`try_into_panic` → `resume_unwind`, preserving inline semantics) is careful and correct.
6. **`evm_database`: `plain_value()` → `find()`**, dropping a heap allocation per state read. A hot-path optimization.

## 2. Public API removals, consumed cross-repo

mega-reth consumes this crate as `stateless-core = { git = ..., tag = "v2.0.14" }`. Removed or changed here:

- `pub fn mega_mainnet_hardforks()`
- `pub const BLOB_GASPRICE_UPDATE_FRACTION` (replaced by revm's `BLOB_BASE_FEE_UPDATE_FRACTION_CANCUN` — same value, 3338477, I checked; worth one confirming line in the body since it is a consensus parameter)
- `pub fn LightWitnessExecutor::kvs()`
- `validate_block` / `validate_block_deriving_updates` / `replay_block` lose the `writer` parameter, which **removes EIP-3155 trace output entirely** — a feature deletion, not a cleanup
- `ChainStore: ContractStore` supertrait dropped

Nothing breaks at the pinned tag, but the next bump has to be coordinated. The supertrait change is a genuine improvement and has a concrete downstream payoff worth naming: mega-reth's `MegaValidatorChainStore` carries a `ContractStore` impl its own comment calls "intentionally inert", existing only to satisfy that bound — it can be deleted on the bump. Please say in the body that the next tag is breaking and what the downstream change is.

## 3. An invariant that lost its only enforcement

Removing `mega_mainnet_hardforks()` and the explicit reordering loop makes `MegaethGenesisHardforks::into_vec`'s literal the single source of activation order. That is the right direction — two sources of the same ordering was the drift risk.

But **nothing tests the order.** `test_merge_mega_hardforks_in_op_hardforks` asserts `spec.hardforks.fork(MegaHardfork::MiniRex) == Timestamp(3)` and similar — map lookups, order-independent. Meanwhile `chain_spec.hardfork(timestamp)` — which #171 uses to select `BlockLimits` — is order-sensitive. A wrong literal order yields wrong block limits, i.e. consensus divergence, with the entire suite green.

Please add a test asserting the relative order of the MegaETH forks in `forks_iter()`, plus a comment on the literal saying new hardforks must be inserted at their activation position and the test bumped alongside.

## Worth keeping, whatever the split

`ChainStore`/`ContractStore` decoupling with its stated rationale; `BackoffSchedule` collecting jitter/cap/1ms-floor into one type (previously duplicated between the R2 and RPC retry loops, comments and all); dropping `in_flight_blocks` as a second source of truth for `task_to_block`; the dead-parameter chain cleanup; replacing the test-only `set_anchor_block` backdoor with the production `reset_to_anchor` path; `BlockMeta::from_header` collapsing four identical field copies.

flyq:
Closing per review: this will be resplit into a behavior-preserving cleanup plus individually-motivated PRs (plan being drawn up; work starts after the currently open PRs merge). The branch stays as source material.

flyq:
Split as agreed, six stacked PRs: #194 (behavior-preserving cleanup + unused-API removals) → #195 (EIP-3155 writer deletion) → #196 (verify_block_integrity / gas_used) → #197 (blocking work off the runtime, incl. the per-attempt-timeout interaction this review flagged) → #198 (hot-path allocations) → #199 (empty witness endpoints + typed range error). Each carries the part of this review it answers.

PR #172, opened 2026-08-01
feat: add an HTTP connect timeout so unreachable RPC hosts fail fast

flyq:
## Summary
`RpcClient` now builds one shared reqwest client with a connect-phase bound (`RpcClientConfig::connect_timeout`, default 3s) and injects it into every alloy provider (data, witness, report). An endpoint whose TCP handshake gets no reply — a dead host, or a firewall silently dropping SYNs — now fails and rotates within ~3s instead of consuming the full 20s `per_attempt_timeout` on every attempt.

## Motivation
Observed on a mainnet debug-trace-server deploy: the internal witness generator endpoint sat behind a firewall that silently dropped SYNs, and every chain-sync block burned the whole 20s per-attempt budget on provider 0 before rotating to the working fallback (which then answered in ~60ms), capping sync throughput at ~0.1 blocks/s. Connection refusal (RST) already fails in milliseconds; only the no-reply case had no bound tighter than `per_attempt_timeout` — and that one cannot simply be lowered without also cutting the budget for legitimately slow multi-MB witness transfers. The two failure modes need separate knobs.

## Change
- New `RpcClientConfig::connect_timeout` field (default 3s, enough for a WAN handshake that loses its first SYN) wired through a single shared `reqwest::Client` (`crates/stateless-common/src/rpc_client.rs:299`); connection pools are keyed per host, so sharing one client across providers is behavior-neutral.
- Providers are now constructed via `alloy_rpc_client::ClientBuilder::http_with_client` + `RootProvider::new` instead of `ProviderBuilder::connect_http`; the alloy builder path preserves the `is_local` guessing of the previous shape.
- New workspace dependency `alloy-rpc-client` (same 1.0.23 family as the existing alloy deps, `default-features = false`; its `reqwest` feature is already enabled through alloy-provider via feature unification). `Cargo.lock` gains only the new dependency edge.

## Testing
New `test_connect_timeout_rotates_past_unreachable_provider`: a backlog-1 listener with a pre-filled accept queue models the SYN-drop case; with `connect_timeout=200ms` and `per_attempt_timeout=60s`, `get_latest_block_number` rotates to the healthy second endpoint well inside a 5s outer bound (the test self-skips if the host kernel refuses to saturate the queue). Negative check performed locally: lifting the connect bound to 50s makes the test fail via the outer timeout, so the simulation genuinely hangs the connect phase rather than getting an RST. `cargo test -p stateless-common`: 41 passed; fmt/clippy/cargo-sort clean.

## Equivalence notes (review follow-up)

- `ProviderBuilder::default()` (unlike `::new()`) carries no fillers, so `RootProvider::new(ClientBuilder::default().http_with_client(..))` is transport-equivalent — the result is still a bare `RootProvider`.
- reqwest enables gzip/brotli automatically when those features are compiled (`Accepts::default()`), so constructing the client explicitly does not drop response compression on multi-MB witness fetches; the `Cargo.toml` comment documents the same unification.
- `connect_timeout` as a CLI/env knob: agreed it is worth exposing (3s is reasonable regionally, tight cross-continent) — ledgered as a follow-up, since it needs CLI/env wiring in both binaries.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5656,6 +5656,7 @@ dependencies = [
  "alloy-consensus",
  "alloy-primitives",
  "alloy-provider",
+ "alloy-rpc-client",
  "alloy-rpc-types-eth",
  "alloy-trie",
  "base64",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ alloy-provider = { version = "1.0.23", features = [
     "reqwest-rustls-tls",
 ], default-features = false }
 alloy-rlp = { version = "0.3.10", default-features = false }
+alloy-rpc-client = { version = "1.0.23", default-features = false }
 alloy-rpc-types-eth = { version = "1.0.23", default-features = false }
 alloy-rpc-types-trace = "1.0.23"
 alloy-serde = { version = "1.0.23", default-features = false }
--- a/crates/stateless-common/Cargo.toml
+++ b/crates/stateless-common/Cargo.toml
@@ -12,6 +12,7 @@ exclude.workspace = true
 alloy-consensus.workspace = true
 alloy-primitives.workspace = true
 alloy-provider.workspace = true
+alloy-rpc-client.workspace = true
 alloy-rpc-types-eth.workspace = true
 alloy-trie.workspace = true
 
@@ -35,7 +36,7 @@ clap.workspace = true
 eyre.workspace = true
 fastrand = { workspace = true, features = ["std"] }
 futures.workspace = true
-# Enables gzip/brotli on alloy-provider's reqwest 0.12 (Cargo feature unification) for witness/data fetches; not referenced in code.
+# Also enables gzip/brotli on the shared reqwest 0.12 client (Cargo feature unification) for witness/data fetches.
 reqwest = { workspace = true, features = ["gzip", "brotli"] }
 rolling-file.workspace = true
 serde.workspace = true
--- a/crates/stateless-common/src/rpc_client.rs
+++ b/crates/stateless-common/src/rpc_client.rs
@@ -2182,11 +2182,12 @@ mod tests {
                 // that is a hard failure: this is the only regression test of
                 // `connect_timeout`, and a silent skip would let a regression ship behind a
                 // green check. Other dev platforms keep the loud skip.
-                assert!(
-                    !cfg!(target_os = "linux"),
-                    "accept queue did not saturate after 64 connects — \
-                     the connect-timeout path was never exercised"
-                );
+                if cfg!(target_os = "linux") {
+                    panic!(
+                        "accept queue did not saturate after 64 connects — \
+                         the connect-timeout path was never exercised"
+                    );
+                }
                 eprintln!("skipping: accept queue did not saturate");
                 return;
             }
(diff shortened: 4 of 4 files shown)


vincent-k2026 (commented):
The design is right — one test issue worth fixing before merge.

**What's good**

The two-tier budget is well reasoned and the reasoning is written down: a separate connect bound rotates a dead host away in seconds, while an established connection keeps the full per-attempt budget for multi-MB transfers. The 3s default is argued from SYN retransmission timing (~1s) rather than picked. Sharing one `reqwest::Client` across providers is also the right call — pools are keyed per host, so data/witness/report endpoints all inherit the bound.

**1. The test can pass without testing anything**

```rust
if queue_filler.len() >= 64 {
    eprintln!("skipping: accept queue did not saturate");
    return;
}
```

This is the only regression test in the PR. If a CI host's kernel accepts more connections than expected (`somaxconn`, container network stack), it returns green having verified nothing — and nothing in CI would tell you that happened.

The test itself discriminates fine: drop `connect_timeout` from the client and the outer 5s timeout trips the `expect`. The problem is only that it has a silent exit. Either make the skip a hard failure on Linux CI, or record a marker CI can assert on. As written, a future regression could ship behind a green check.

**2. Two things worth stating in the description**

Both are load-bearing to the change being a no-op outside the timeout, and both are the first things a reviewer will want to confirm:

- `ProviderBuilder::default()` (as opposed to `::new()`) carries no fillers, so replacing it with `RootProvider::new(ClientBuilder::default().http_with_client(..))` is transport-equivalent — the result is still a bare `RootProvider`.
- reqwest enables gzip/brotli automatically when those features are on (`Accepts::default()`), so constructing the client explicitly does not silently drop response compression on multi-MB witness fetches. The `Cargo.toml` comment was updated correctly for this, but the body should say it.

**3. Question**

Should `connect_timeout` be exposed as a CLI/env knob alongside the other `RpcClientConfig` fields? 3s is reasonable for regional deployments but tight for cross-continent ones, and this is precisely the value an operator wants to move during a partial network incident.

flyq:
@vincent-k2026 Test hardened in 4f2f069: on Linux — the CI platform — failure to saturate the accept queue is now a hard assert, so the only regression test of `connect_timeout` can no longer pass silently; other dev platforms keep the loud skip. Both equivalence statements are now in the description (`ProviderBuilder::default()` carries no fillers; the explicit client keeps reqwest's gzip/brotli). On the knob question: agreed it's worth exposing — ledgered as a follow-up since it needs CLI/env wiring in both binaries, keeping this PR minimal.

vincent-k2026 (approved):
Re-reviewed at `44ef414`. The one blocking item is closed — approving.

**The silent skip is gone.** On Linux — the CI platform — failing to saturate the accept queue after 64 connects now panics instead of returning green, so the only regression test of `connect_timeout` can no longer pass without having exercised the connect path. Other platforms keep the loud skip, which is the right split: the guarantee is enforced where it is actually checked, not weakened everywhere to accommodate dev laptops. I read the whole test at HEAD; that `>= 64` branch was the only way out that did not run an assertion.

The clippy follow-up in `44ef414` is the correct shape too — `cfg!` expands to a constant so `assert!` tripped `assertions_on_constants`; `if cfg!(...) { panic!(...) }` keeps the semantics without suppressing the lint.

Both equivalence statements are in the body now (`ProviderBuilder::default()` carries no fillers, reqwest keeps gzip/brotli via `Accepts::default()`), which is what a future reader will need to convince themselves the transport swap was behaviour-neutral. Ledgering the `connect_timeout` CLI/env knob as a follow-up is fine — it needs wiring in both binaries and does not belong in this PR.

`label-check` is no longer red. `pr-review` is red on every open PR in the repo (org `claude-pr-review` action fails with `ENOENT ... symlink` before reviewing) — infra, not this PR.

Troublor (approved):

PR #174, opened 2026-08-03
perf(debug-trace-server): negotiate gzip/zstd response compression at the origin

flyq:
## Summary
Adds negotiated HTTP response compression to debug-trace-server: clients that send `Accept-Encoding: gzip` (or `zstd`) get compressed bodies, everyone else keeps byte-identical identity responses — JSON-RPC payloads are untouched, this is plain RFC 9110 content negotiation. Bodies ≤ 4 KiB always stay identity. Ops kill switch: `--response-compression-disabled` / `DEBUG_TRACE_SERVER_RESPONSE_COMPRESSION_DISABLED`.

## Motivation
Monster-block traces serve 18–21 MB of callTracer JSON per block, and from a user's vantage total latency is transfer-dominated (~0.4s TTFB + 3–6s body transfer through the public gateway). The fronting CDN (Cloudflare) already gzips edge→client when asked — measured 20.47 MB → 3.41 MB (6.0×) on a mainnet monster block — but every POST is a cache MISS that goes back to origin uncompressed, so edge compression barely moves end-to-end time (31s → 27s from one vantage). Compressing at the origin shrinks the back-to-origin leg too, and trace JSON is repetitive enough that the fastest level captures most of the win.

## Implementation
- `compression::layer(enabled)` (`bin/debug-trace-server/src/compression.rs:31`) wraps tower-http's `CompressionLayer`, pinned to `CompressionLevel::Fastest`, gzip+zstd only (no br: CPU-heavier, and gzip+zstd cover browsers/curl/SDKs). A `SizeAbove(4096)` predicate (`compression.rs:18`) keeps tiny JSON-RPC responses identity — compressing them would cost a fresh encoder allocation and their `Content-Length` for a few dozen saved bytes.
- The middleware stack is composed in exactly one place, `http_middleware()` (`bin/debug-trace-server/src/main.rs:425`), used by both the server and the unit tests — so the ordering invariant is executable, not prose: compression outermost keeps `x-response-size` reporting the uncompressed payload size (`ResponseSizeLayer` reads the body's exact `size_hint`, gone once compressed) and keeps `x-execution-time-ns` free of compression CPU (the timing future resolves before the body streams).
- Negative flag mirrors the existing `--response-cache-disabled` style (`bin/debug-trace-server/src/main.rs:193`), reported in the `"Server configuration"` debug log next to its siblings, with default/CLI/env parse coverage (`main.rs:1156`).
- Docs: README "Response compression" block, AGENTS.md one-liner.

## Validation on tko (2026-08-03, combined with #167)
Deployed together with #167 on the tko staging box and benched on the 51-block monster segment (callTracer, 20 MB/block, response cache warm; full data in `validator-data/.../bench/0803/`, analysis in `bench/old-vs-new-report-2026-08-01.md` §9):

| warm rounds (51 hits) | identity | gzip | zstd |
| --- | --- | --- | --- |
| total p50 | 125–133 ms | **96–101 ms** | **62–71 ms** |
| wire bytes per block | 20.0 MB | **3.46 MB (5.8×)** | **2.48 MB (8.1×)** |
| server CPU p50 (header) | 12 ms | 12 ms | 12 ms |

Compression is a net latency win even on loopback: it pipelines with hyper's write loop and the saved socket bytes outweigh the compression time. Real CPU cost estimated from round walls: gzip ≈ +47 ms/block, zstd ≈ 0–20 ms/block — far below the pre-measurement estimate, so the per-encoding compressed-variant cache idea is downgraded to a watch item. Zero regressions vs the 2026-08-01 baseline (`--compare`: 0 regressed, 51/51 OK everywhere); response-cache hit counters unaffected by encoding.

## Testing
- CI unit tests drive requests through the real production stack (`crate::http_middleware`) against a mock JSON service (`bin/debug-trace-server/src/compression.rs:40` onward): gzip and zstd negotiation with round-trip decode, identity passthrough without `Accept-Encoding`, sub-threshold responses staying identity, kill-switch passthrough, and `x-response-size`/`x-execution-time-ns` surviving compression.
- Live `#[ignore]` integration test `bin/debug-trace-server/tests/compression_test.rs` mirrors timing_header_test.rs against a running server. Cargo feature unification turns on reqwest's gzip/brotli here (via stateless-common), so the client opts out explicitly with `no_gzip()/no_brotli()` to keep raw-byte assertions valid.
- `cargo test -p debug-trace-server --bin debug-trace-server`: 112 passed; fmt/clippy/cargo-sort clean.

## Notes
- Compressed responses switch to chunked transfer (no `Content-Length`); Cloudflare and the bench tooling handle this fine, and CF will now pass through the origin's `content-encoding` instead of re-compressing at the edge.
- Response-cache hits re-compress per hit by design (the cache stores identity JSON); measured cheap enough above that this is documented behavior, not a follow-up blocker.
- Cargo.lock churn is the tower-http/flate2 graph additions only (cargo-managed).
- BREACH/CRIME does not apply: those attacks need attacker-controlled input compressed in the same context as a secret, and trace responses are public chain data with no per-client secret mixed in — there is no compression oracle to exploit on this endpoint.


bin/debug-trace-server/src/compression.rs
```diff
@@ -0,0 +1,39 @@
...
+/// where the fast level already collapses most of the redundancy while keeping
+/// per-response CPU small next to the trace itself.
+pub(crate) fn layer(enabled: bool) -> ResponseCompressionLayer {
+    // size gate first: it short-circuits the content-type checks for the small
+    // responses it exists to reject
+    CompressionLayer::new()
+        .quality(CompressionLevel::Fastest)
+        .gzip(enabled)
+        .zstd(enabled)
+        .compress_when(SizeAbove::new(MIN_COMPRESS_SIZE).and(DefaultPredicate::new()))
```
Troublor, line 46: The kill switch only covers the encodings that happen to be compiled in. `CompressionLayer::new()` starts from `AcceptEncoding` with **all four** encodings enabled (`compression_utils.rs`, `impl Default for AcceptEncoding`); this overrides `gzip` and `zstd` only.

tower-http's features are additive across the workspace graph, so the moment anything pulls in `compression-br` or `compression-deflate` (directly or via `compression-full`), brotli becomes negotiable — and browsers advertise `br` by default. Two things then break silently at once: the doc claim just above ("`enabled = false` turns every negotiable encoding off") stops being true, so an operator flipping `--response-compression-disabled` mid-incident sees nothing change; and the deliberate "gzip+zstd only, no br — CPU-heavier" policy widens with no code change and no failing test.

This isn't hypothetical for this repo — `crates/stateless-common/Cargo.toml:38-39` turns on reqwest's `gzip`/`brotli` purely through unification, and `compression_test.rs:19-22` documents working around that exact effect.

`no_br()` / `no_deflate()` are **not** `#[cfg]`-gated (unlike `br()`/`deflate()`), so they compile today regardless of features and pin the negotiated set to what the module claims:

```suggestion
        .gzip(enabled)
        .zstd(enabled)
        .no_br()
        .no_deflate()
        .compress_when(SizeAbove::new(MIN_COMPRESS_SIZE).and(DefaultPredicate::new()))
```

Worth pinning with a test through `http_middleware(true)` sending `Accept-Encoding: br, deflate` and asserting an identity response.

bin/debug-trace-server/src/middleware.rs
```diff
@@ -0,0 +1,156 @@
+//! Composes the HTTP middleware stack — the one place its order is defined.
+//!
+//! Compression must stay outermost: `ResponseSizeLayer` reads the body's exact
+//! `size_hint` (gone once the body is a compressed stream), and the timing layer's
+//! future resolves before the body streams — so this order keeps `x-response-size` at
+//! the uncompressed payload size and `x-execution-time-ns` free of compression CPU.
```
Troublor: `x-execution-time-ns` being "free of compression CPU" is framed as a benefit here, but it leaves the compression cost accounted for nowhere. The timing future writes the header *and* records the `CpuTimeMetrics` histogram (`timing.rs:117`), `ResponseSizeLayer` reads the identity body's `size_hint`, and `ResponseSizeMetrics` is recorded further down in `rpc_service.rs:398` on the identity JSON — all of them complete before the encoder is ever polled, since it only runs in `WrapBody::poll_frame` during body streaming.

So the ~47 ms/block of gzip CPU measured in the PR description lands in no header and no metric. Concretely: an opted-in caller drives cache-hit-dominated `callTracer` traffic, every hit re-gzips 20 MB (the documented trade-off), the box goes CPU-bound — and `x-execution-time-ns`, `CpuTimeMetrics`, and `ResponseSizeMetrics` all read exactly as before. The server reports itself healthy while clients time out; the only signal is host CPU or client-side latency. That's uncomfortable given `--response-compression-disabled` exists specifically to shed this CPU: nothing scrape-visible tells an operator when to pull it.

Please keep this layer order — it's the right one — and instead measure the compression CPU separately in the body path, folding it into the existing timing statistics. A thin wrapper around the compressed body that samples `CLOCK_THREAD_CPUTIME_ID` around each `poll_frame` (the same measurement `timing.rs` already does) and records the total into `CpuTimeMetrics` on end-of-stream would make the histogram whole again without touching `x-response-size`.

Worth being explicit about one limit: the per-response `x-execution-time-ns` header itself can't carry it, since it's flushed with the response head before any compressed byte exists — HTTP trailers are the only header-shaped option, and most clients won't read them. So the metric is the realistic target, and the README line claiming the header excludes compression CPU should probably say the header no longer bounds total request cost.

While you're in there, a compressed-wire-bytes counter would also be valuable — with `x-response-size` reporting the uncompressed size, wire throughput is currently underivable from server telemetry.

bin/debug-trace-server/src/compression.rs
```diff
@@ -0,0 +1,39 @@
...
+    compression::{
+        CompressionLayer,
+        predicate::{And, DefaultPredicate, Predicate, SizeAbove},
+    },
+};
+
+/// Bodies at or below this size are served identity even for opted-in clients:
+/// compressing them costs a fresh per-response encoder allocation and downgrades
+/// `Content-Length` to chunked framing for a saving of a few dozen wire bytes.
+pub(crate) const MIN_COMPRESS_SIZE: u16 = 4096;
```
Troublor: Off by one against the code. `SizeAbove::should_compress` is `size >= min` (tower-http 0.6.8 `compression/predicate.rs:184`), so a body of *exactly* 4096 bytes **is** compressed — "at or below" is wrong at the boundary.

Three other places repeat the same claim: `README.md:111` ("4 KiB or less"), `AGENTS.md:110` ("bodies ≤ 4 KiB stay identity"), and `tests/compression_test.rs:88-92`, whose skip branch asserts `content-encoding` is absent whenever `x-response-size <= MIN_COMPRESS_SIZE` — that assertion fails on a 4096-byte response, so the live test is flaky by construction at the boundary.

Either reword all four to "below":

```suggestion
/// Bodies below this size are served identity even for opted-in clients:
/// compressing them costs a fresh per-response encoder allocation and downgrades
/// `Content-Length` to chunked framing for a saving of a few dozen wire bytes.
pub(crate) const MIN_COMPRESS_SIZE: u16 = 4096;
```

...and change the test's `<=` to `<`; or set the constant to `4097`, which makes all four existing statements true with a one-line change. Either way, a unit test at exactly `MIN_COMPRESS_SIZE` (and ±1) would pin the boundary — `small_response_stays_identity` currently uses a 38-byte body, so it never exercises it.

README.md
```diff
@@ -106,6 +106,14 @@ JS tracers, `muxTracer`, and struct-logger requests with non-default flags bypas
...
 Disable the cache with `--response-cache-disabled`; `--response-cache-estimated-items` must be at least 1 (the old `=0` disable convention is rejected at startup).
 
+**Response compression:**
+Responses negotiate gzip/zstd per request via the client's `Accept-Encoding` header; clients that do not send it keep receiving identity bodies, so nothing changes for consumers that have not opted in.
+Bodies of 4 KiB or less are always served identity: compressing them would cost a per-response encoder allocation and downgrade `Content-Length` to chunked framing for a few dozen saved bytes.
+Compression runs at the fastest level while the body streams, so `x-execution-time-ns` excludes its CPU cost and `x-response-size` keeps reporting the uncompressed payload size.
+Compressing at the origin also shrinks a fronting CDN's back-to-origin leg, which edge-side compression alone cannot do.
+Known trade-off: the response cache stores identity JSON, so an opted-in client pays the (fast-level) compression on every cache hit.
+Disable with `--response-compression-disabled` if compression CPU needs to be shed at the server immediately.
```
Troublor: "immediately" overstates what this flag can do. It's a `clap` startup flag read once at `main.rs:633`, so using it means restarting the process — which discards the response cache, the `BlockData` cache (1 GB default), and the canonical-hash memo, all in-memory and process-scoped per AGENTS.md.

An operator following this sentence during a CPU incident makes latency *worse* first, for the whole cold-start window. Either soften the wording:

```suggestion
Disable with `--response-compression-disabled`; the flag is read at startup, so shedding compression CPU means a restart and a cold-cache window.
```

or make it runtime-settable by having the compression predicate consult a shared `AtomicBool`.

Troublor (changes requested):

bin/debug-trace-server/src/compression.rs
flyq: Fixed in 28fe7e8 exactly as suggested: `.no_br().no_deflate()` added, with a comment marking them load-bearing against workspace feature unification rather than restated defaults. Also added the pin test you proposed — `br_and_deflate_are_not_negotiated` drives `Accept-Encoding: br, deflate` through `http_middleware(true)` and asserts an identity response. It's trivially green today and becomes the alarm the moment any dependency unifies `compression-br`/`compression-deflate` into the graph. Good catch — the stateless-common/reqwest precedent made this a matter of time, not hypothesis.

bin/debug-trace-server/src/middleware.rs
flyq: Implemented in 28fe7e8 as a new outermost middleware, `body_metrics.rs`: it wraps the compressed response body and samples `CLOCK_THREAD_CPUTIME_ID` around each `poll_frame` (same clock as timing.rs, now shared via `timing::thread_cpu_time`), recording at end-of-stream — or on drop, so a client that aborts mid-transfer still accounts the CPU it burned. It also records the post-compression wire bytes you asked for.

One deliberate deviation from the suggestion: the total goes to a dedicated `debug_trace_body_cpu_time_seconds` histogram plus `debug_trace_wire_bytes_total` counter (both labeled by negotiated encoding) instead of folding into `CpuTimeMetrics` — two observations per request in one summary would muddy the per-request quantiles, and a dedicated series is directly alertable for exactly the "when to pull the kill switch" question. Happy to fold it in instead if you feel strongly.

README now states the header reflects request processing and no longer bounds total request cost, pointing at the new metrics; agreed on trailers being impractical, the header is untouched.

bin/debug-trace-server/src/compression.rs
flyq: Took the rewording option in 28fe7e8: the const doc now reads "below this size", README says "under 4 KiB", AGENTS.md likewise, and the live test's skip branch is `<` with its message updated. Also added the boundary unit test you suggested — `threshold_boundary_is_exact` drives exactly-4096 (compressed) and 4095 (identity) bodies through the real stack.

README.md
flyq: Applied your suggested wording verbatim in 28fe7e8. Left the runtime `AtomicBool` variant out of this PR's scope — with the new `debug_trace_body_cpu_time_seconds` metric (other thread) telling operators when compression CPU matters, we can judge from real data whether the restart cost ever justifies a live toggle.

vincent-k2026 (commented):
Reviewed against HEAD `28fe7e8`. Composing the middleware in one place and pinning the ordering invariant with tests that drive the real `http_middleware()` rather than a replica is the right shape — the invariant is executable instead of prose.

One detail in `body_metrics.rs` worth naming because it is easy to get wrong: `CLOCK_THREAD_CPUTIME_ID` is per-thread and tokio work-steals between polls, but `start` and `end` are both taken inside a single `poll_frame` call, which cannot migrate mid-call — so each delta is valid and summing them across threads is sound. Recording on drop as well as end-of-stream, so an aborted transfer still accounts its CPU, is the right call too.

Four non-blocking items.

## 1. `BodyMetrics` is not pre-registered

`pre_register_all_metrics()` deliberately registers every metric family at startup so series exist at zero rather than being absent — `RpcMethodMetrics`, `ResponseSizeMetrics`, `CpuTimeMetrics`, `CacheMetrics` are all in there. `BodyMetrics::create(encoding)` only creates its series when the first response with that encoding is served.

So `debug_trace_wire_bytes_total{encoding="gzip"}` is **absent**, not zero, until the first gzip request. An alert or panel built on it is broken during exactly the window before traffic shows up. Three lines fixes it:

```rust
// Body streaming (per negotiated encoding)
for enc in ["identity", "gzip", "zstd"] {
    let _ = BodyMetrics::create(enc);
}
```

## 2. `debug_trace_body_cpu_time_seconds` has no `BUCKET_SPECS` entry, so it renders as a summary

`metrics-exporter-prometheus` emits histograms without a configured bucket set as **summaries** — per-instance quantiles, which do not aggregate across replicas. That collides with the stated reason for making this a dedicated series: it is meant to be directly alertable for the "when do I pull the kill switch" question, and you cannot cleanly alert on a summary across a multi-replica deployment.

Worth an explicit bucket spec — and arguably for the existing `CpuTimeMetrics` histogram too, which has the same gap, so this is an inherited shape rather than something this PR introduced.

## 3. Boundary inconsistency in the live test

`tests/compression_test.rs:847`:

```rust
assert!(body.len() <= MIN_COMPRESS_SIZE, "cache status unexpectedly large: {}", body.len());
```

`SizeAbove::should_compress` is `size >= min`, so a `debug_getCacheStatus` body of exactly 4096 bytes *is* compressed and the `content-encoding` assertion on the next line would fail. Very unlikely to trigger, but it is a one-character change to bring this in line with how the threshold is described everywhere else.

## 4. Worth a sentence in the description

Enabling compression on a public unauthenticated endpoint invites the BREACH/CRIME question. It does not apply here — trace responses carry no secret to leak — but stating that explicitly is cheaper than answering it later, especially for origin-side compression on a CDN-fronted endpoint.

flyq:
@vincent-k2026 Items 1–3 fixed in 41c1613: `BodyMetrics` pre-registered for identity/gzip/zstd (the only three encodings the layer can ever negotiate); `debug_trace_body_cpu_time_seconds` got an explicit `BUCKET_SPECS` entry so it renders as an aggregatable histogram; the live-test boundary assert is now `<`. One deliberate non-change: the existing `cpu_time_seconds` keeps its summary shape for now — converting it would break dashboards already reading its quantile series, so that migration should be its own coordinated change. Item 4: the BREACH/CRIME non-applicability statement is in the description. Thanks for writing down the per-thread-clock soundness argument — that reasoning now has a home in the review record.

vincent-k2026 (approved):

Troublor (approved):

PR #175, opened 2026-08-06
feat(debug-trace-server): direct-from-R2 historical witnesses

flyq:
## Summary
Fixes the 2026-08-05 incident class where bulk backfill traffic (an indexer flying 15 parallel JSON-RPC batches of 20 `debug_traceBlockByNumber`) pushed historical witness fetches on the shared public RPC gateway past the witness budget, surfacing bursts of `-32001 witness fetch exceeded deadline` to the client. Historical witnesses now come straight from the R2 bucket — which holds full history and tolerates far higher parallelism — with the RPC chain as fallback. The feature is off by default (activates only when the `--r2-*` flag group is configured), so the unflagged binary reproduces current production behavior exactly (useful for A/B benchmarking). A sequential-readahead prefetcher was part of earlier revisions and has been dropped: fra benchmarks showed R2 direct reads alone sustain ~2x the incident client's consumption rate, and a request-triggered prefetch fan-out is an amplification surface.

## Changes
- **`stateless-r2`: shared witness-object GET fetcher** (`crates/stateless-r2/src/fetch.rs`). The SigV4-signed GET, response classification (`R2GetError`), jittered-backoff retry loop, and concurrency cap move out of the validator's `r2_witness.rs` into `stateless_r2::fetch::R2ObjectFetcher`, shared by both binaries instead of copied. The loop is now deadline-aware: each attempt's HTTP timeout clamps to the remaining budget, and no retry starts whose backoff sleep would overrun it (pipeline callers pass `None` and keep the attempt-cap behavior). The validator keeps a thin adapter — full decode (proof verification needs the EC points), its metrics, and the pipeline's surfaced-failure pacing — with unchanged behavior; the scripted mock R2 server moves to `stateless-test-utils` for reuse by all three test suites.
- **debug-trace-server: direct-from-R2 historical witnesses** (`bin/debug-trace-server/src/r2_witness.rs`, wired in `fetch_witness` at `bin/debug-trace-server/src/data_provider.rs:1237`). With the all-or-nothing `--r2-endpoint` / `--r2-bucket` / `--r2-access-key-id` / `--r2-secret-access-key` group (env `DEBUG_TRACE_SERVER_R2_*`; clap `requires` enforces pairing), blocks at least `--witness-local-window` below the local tip try a signed R2 GET before the RPC chain. Light decode only (the trace server never verifies proofs, so the full decode's per-point EC work buys nothing). The R2 attempt gets half the remaining witness budget so a hung endpoint can never starve the RPC fallback, retries at most 3 times, and surfaces failures with zero pacing pauses — any failure (missing object, throttle, transport, corrupt payload) falls back to the RPC chain immediately. Frontier blocks keep the generator path (the bucket lags the uploader there). `--r2-max-concurrent-requests` caps R2 GETs separately from the RPC witness semaphore, which is sized for a shared gateway.
- **Metrics**: witness source label `witness_r2`, `debug_trace_r2_witness_retries_total`, `debug_trace_r2_witness_errors_total{kind}`, all pre-registered.
- `RedactedSecret` moves from the validator's `app.rs` to `stateless-common` so both binaries' secret flags share it.

## Testing
- debug-trace-server: 125 unit tests, including new ones for R2-first routing / R2-miss RPC fallback / recent-blocks-never-touch-R2, CLI parsing of the all-or-nothing `--r2-*` group, and KINDS-superset drift guards tying both adapters' metric pre-registration to `R2GetError::KINDS`.
- stateless-r2: 32 tests covering classification, retry pacing, deadline-stops-retries, expired-deadline fast-fail, `on_retry` counting, and the concurrency cap; validator suite (25 lib + 13 integration) unchanged and green, pinning the refactor as behavior-preserving.
- `cargo clippy --workspace --all-targets --all-features` zero warnings; `cargo fmt` / `cargo sort` clean; `stateless-core --no-default-features` lib compile checked.

## Notes
- `stateless-test-utils` now declares `revm = { features = ["serde"] }` itself: it always needed it and previously leaned on sibling crates' feature unification, which broke the moment it was built standalone from stateless-r2's dev-deps.
- `stateless-r2` gains tokio/tracing/fastrand dependencies for the fetcher; reqwest stays pinned at 0.12 (API-visible to mega-reth's workspace).
- A genuinely missing R2 object (`NoSuchKey`) is deterministic — not retried, silent fallback to the RPC path; watch `debug_trace_r2_witness_errors_total{kind="missing"}` for bucket completeness gaps.
- Deployment guidance from the fra witness-path benchmark (300-witness waves against the incident block range): set `DEBUG_TRACE_SERVER_R2_MAX_CONCURRENT_REQUESTS=48`. The R2 client is HTTP/1.1 (one connection per in-flight GET), and the egress IP's concurrent-connection budget at the Cloudflare edge is adaptive — baseline ~128, but connection storms lower it below 64 for 30+ minutes; 48 measured clean at 144 wit/s (~2x the incident client's consumption) with per-GET p50 ~284ms (pure fra→APAC backbone RTT); with the fast hang-detection below, running somewhat above (e.g. 56) is an acceptable operator choice. The mitigation's failure mode is handshakes that hang without erroring, so connection establishment is bounded by `--r2-connect-timeout-ms` (default 1000 — a healthy handshake to the local anycast edge is tens of ms) and connect-phase failures surface as their own retryable `kind="connect"` error: a spike in `debug_trace_r2_witness_errors_total{kind="connect"}` is the operational signal that the connection-budget mitigation is active, and each affected request falls back to the RPC chain within ~1s instead of bleeding its witness budget. HTTP/2 cannot lift this: R2's S3 endpoint negotiates http/1.1 only (ALPN-verified by the new `stateless-r2` example `r2_probe`, which also serves as a deploy-time credentials/wiring check), so connection count is intrinsically tied to concurrency on that endpoint. The lever beyond 48 is an R2 custom domain (rides the regular CDN stack: negotiates h2, and can edge-cache the immutable witness objects) — ops setup pending, client support in a follow-up PR.
- Today's production fallback shape (single `mega_getBlockWitness` calls at 20 concurrent) measured ~66 wit/s — approximately the incident client's consumption rate, which is why the 0805 timeouts occurred.


bin/debug-trace-server/src/data_provider.rs
```diff
@@ -1150,6 +1184,49 @@ async fn fetch_witness(
...
+/// One R2 attempt for a historical witness, on [`R2_WITNESS_BUDGET_DIVISOR`]'s share of the
+/// remaining budget. `None` on any failure — the caller falls back to the RPC chain.
+async fn try_r2_witness(
+    r2: &R2WitnessSource,
+    block_number: u64,
+    block_hash: B256,
+    deadline: Instant,
+) -> Option<(LightWitness, MptWitness)> {
+    let now = Instant::now();
+    let r2_deadline = now + deadline.saturating_duration_since(now) / R2_WITNESS_BUDGET_DIVISOR;
```
vincent-k2026, line 1196: **Blocking.** This divisor is the one safety invariant the PR rests on — the const's own comment says "a hung R2 endpoint can never starve the RPC fallback" — and nothing tests it. Delete `/ R2_WITNESS_BUDGET_DIVISOR` and all three new R2 tests still pass.

`fetch_witness_falls_back_to_the_rpc_chain_when_r2_misses` uses a 404, which returns instantly, so it pins the *fallback wiring*, not the *budget split*. A PR whose entire justification is protecting the witness budget should have the protection itself under test.

`stateless_test_utils::mock_r2::mock_r2_held` is already there: point R2 at a held endpoint that never answers, give the RPC mock a working witness, and assert `fetch_witness` still returns inside the outer deadline. That test must go red if the divisor is removed or set to 1.

bin/debug-trace-server/src/main.rs
```diff
@@ -494,6 +539,50 @@ async fn main() -> Result<()> {
...
+        // can never fire — skip building the client instead of holding a dead one.
+        (Some(_), ..) if args.data_dir.is_none() => {
+            warn!(
+                "--r2-endpoint without --data-dir: block age cannot be anchored to a \
+                 local tip, so the R2 historical route is inactive and all witnesses \
+                 use the RPC chain"
+            );
+            None
+        }
+        (Some(endpoint), Some(bucket), Some(access_key_id), Some(secret)) => {
```
vincent-k2026, line 579: **Blocking.** This arm only checks presence, not emptiness, so `DEBUG_TRACE_SERVER_R2_BUCKET=""` (a secret that failed to inject is the common shape) builds a live `R2WitnessSource` over an empty bucket.

The consequence is worse than a wasted GET: the signed GET 404s with no `<Code>` element, `s3_error_code` returns `None`, and `fetch.rs`'s rule classifies it as **`Missing`**. So a misconfiguration lands on `debug_trace_r2_witness_errors_total{kind="missing"}` — the exact counter this PR's Notes tell operators to watch for *bucket completeness gaps*. A bad deploy will read as a data gap and send someone to audit the bucket.

The validator already gets this right: `require_r2` has `.filter(|v| !v.is_empty())` with a test asserting "an env var set to the empty string must not pass as configured". Same treatment here — one filter.

bin/debug-trace-server/src/main.rs
```diff
@@ -494,6 +539,50 @@ async fn main() -> Result<()> {
...
+    // "absent". Shares the RPC path's per-attempt timeout and retry pacing.
+    let r2_witness_source = match (
+        &args.r2_endpoint,
+        &args.r2_bucket,
+        &args.r2_access_key_id,
+        &args.r2_secret_access_key,
+    ) {
+        // Without a local DB there is no tip to anchor block age, so the historical route
+        // can never fire — skip building the client instead of holding a dead one.
+        (Some(_), ..) if args.data_dir.is_none() => {
```
vincent-k2026: Same class as the empty-value arm above, and I'd make it an error rather than a warn. An operator who passed `--r2-endpoint` plus credentials and no `--data-dir` has explicitly asked for the R2 route and silently gets the exact configuration that produced the 0805 incident — with one `warn!` in a busy startup log as the only signal.

The validator's `require_r2` fails closed for the analogous mistake. Failing startup here is consistent and costs nothing: the flag combination is unusable by construction, not a degraded mode worth running.

bin/debug-trace-server/src/r2_witness.rs
```diff
@@ -0,0 +1,239 @@
...
+            .fetcher
+            .get_block_object(
+                number,
+                hash,
+                MAX_ATTEMPTS,
+                Some(deadline),
+                metrics::record_r2_witness_retry,
+            )
+            .await?;
+        let bytes = fetched.bytes;
```
vincent-k2026, line 132: Non-blocking, but I want to know if it's deliberate: `fetched.queue_wait` is dropped here. `FetchedObject`'s doc says it is reported separately "so duration metrics can subtract it: queue wait on the self-imposed cap folded into a fetch-duration metric would masquerade as R2 slowness", and the validator adapter does subtract it — this consumer doesn't.

It matters most exactly here: the deployment guidance recommends `--r2-max-concurrent-requests=48` for *this* binary, so queueing is the expected steady state, and `witness_r2`'s duration histogram will report self-imposed queue time as R2 latency.

There's a defensible argument the other way — on a request path the user really did wait in that queue, so end-to-end is the honest number, unlike the validator's throughput pipeline. If that's the intent, fine, but then say so on the metric: right now the shared abstraction documents one contract and one of its two consumers silently does the opposite.

bin/debug-trace-server/src/r2_witness.rs
```diff
@@ -0,0 +1,239 @@
...
+    Decode { number: u64, key: String, source: WitnessDecodingError },
+    /// The decode task panicked. This is a bug in our own decoder, not a problem with the
+    /// data in R2, so it is kept out of [`Self::Decode`].
+    #[error("R2 witness decode task for block {number} (key {key}) panicked: {source}")]
+    DecodePanicked { number: u64, key: String, source: JoinError },
+}
+
+impl R2WitnessError {
+    /// Every label [`Self::kind`] can produce, for metrics pre-registration.
+    pub const KINDS: &'static [&'static str] = &[
```
vincent-k2026, line 52: Non-blocking / follow-up-shaped. `ea170f9` deduped the adapter↔fetcher axis, but the adapter↔adapter axis is now a verbatim copy: this `R2WitnessError` and the validator's have the same three variants with the same source types (both decode errors are `WitnessDecodingError`), the same eight-element `KINDS` in the same order, the same `kind()`, and the same `spawn_blocking` decode match. Even `kinds_cover_all_fetch_kinds` exists twice.

That last one is the tell — two identical tests written to *detect* drift in a surface that could just be removed. The error type alone needs no generics to hoist (both use `WitnessDecodingError`), which deletes ~50 lines and both guard tests; the two adapters' real differences are only the decode fn, the metrics fns, and the pacing pause.

A symptom of the copy while it stands: the validator's `KINDS` carries `"deadline"`, but it passes `deadline: None`, so `R2GetError::Deadline` is structurally unreachable there — a permanently-zero series pre-registered on the validator's dashboard.

Happy for this to be a follow-up if you'd rather not widen the PR.

vincent-k2026 (changes requested):
Reviewed against HEAD `59ab2cd`. The layering is right — transport (signed GET, classification, backoff, concurrency cap) in `stateless-r2`, policy (decode depth, attempt budget, pacing, metrics) in each adapter — and the R2-first / RPC-fallback / half-budget shape matches the incident it targets. Two things I'd like fixed before merge, both cheap, plus three smaller notes.

Specific things I liked: `concurrency_limit_bounds_in_flight_gets` asserts `peak == LIMIT` as a liveness guard, not just `peak <= LIMIT` — without it the test could pass having never reached the cap. The two deadline tests genuinely discriminate (drop the deadline-aware acquire and `queued_fetch_surfaces_deadline_before_permit_frees` goes red). And `e37c846` triaging the eight Codex findings in the commit message — including *declining* the decode-deadline one with a stated reason (`spawn_blocking` isn't cancellable, so abandoning the wait saves no CPU) — is the right way to close a bot finding. I reached the same conclusion independently.

Two PR-body fixes while you're in here: the Notes section cites `stateless-r2`'s `r2_probe` example as the ALPN evidence, but `d40ef2c` parked it out of tree (zero hits repo-wide), so the h2 conclusion now points at nothing; and "the validator keeps a thin adapter ... with unchanged behavior" isn't quite true — the validator now gets a 1s connect timeout it didn't have before (`Client::builder()` previously set only `.timeout()`).

bin/debug-trace-server/src/data_provider.rs
flyq: Fixed in 0c6d9f3 — `fetch_witness_hung_r2_leaves_budget_for_the_rpc_fallback` is exactly the test you outlined: `mock_r2_held` R2 that accepts and never answers, a healthy RPC witness mock behind it, and assertions that the fetch succeeds inside the original deadline while R2 was held to (at least) its half share before handing over.

Verified the teeth both ways: with `R2_WITNESS_BUDGET_DIVISOR = 1` the test fails (R2 holds the whole deadline and the fallback gets nothing), at 2 it passes.

bin/debug-trace-server/src/main.rs
flyq: Fixed in 0c6d9f3. `validate_args` now rejects an empty value on any of the four `--r2-*` flags before anything is built — same fail-closed outcome as the validator's `require_r2`, with the error message naming the empty-env-injection shape. One test per flag.

Good catch on the `kind="missing"` consequence in particular; keeping that counter trustworthy as the bucket-completeness signal is exactly why this needed to fail loud rather than filter quietly.

bin/debug-trace-server/src/main.rs
flyq: Agreed — fixed in 0c6d9f3. `--r2-endpoint` without `--data-dir` is now a startup error in `validate_args`, and the construction's warn arm is gone (unreachable after validation). "Unusable by construction, not a degraded mode worth running" settled it.

bin/debug-trace-server/src/r2_witness.rs
flyq: It was deliberate, but you're right that the divergence from the shared contract deserved to be explicit — and a comment alone wouldn't help an operator tuning the cap. 0c6d9f3 does both: the `witness_r2` duration stays end-to-end (on the request path the caller really did wait through that queue), and the queued share now lands on its own pre-registered `debug_trace_r2_witness_queue_wait_seconds` histogram, with the intent documented at the adapter. That histogram is also the signal we'll watch when raising `--r2-max-concurrent-requests`.

bin/debug-trace-server/src/r2_witness.rs
flyq: Taking the follow-up offer — logged with analysis, planned as a small PR after this one merges.

One wrinkle worth recording now: the natural hoist target is not `stateless-r2`. The shared error type needs `WitnessDecodingError` from `stateless-common`, and `stateless-r2` is a leaf consumed by mega-reth's uploaders — pulling the RpcClient stack into that build graph is a real cost. The workable direction is hosting the shared adapter error in `stateless-common` (a common → r2 edge is acyclic; r2 deliberately does not depend on common today). The validator's structurally-unreachable `deadline` kind — nice catch — gets resolved in the same pass: shared `KINDS` makes it definitionally covered, or we feed the validator a real deadline.

vincent-k2026 (approved):
Re-reviewed against HEAD `0c6d9f3`. All five items from my round are closed, and the round-2 commit is tightly scoped — 4 files, +139/-12, nothing else rode along. Approving.

**1. Budget split (was blocking).** `fetch_witness_hung_r2_leaves_budget_for_the_rpc_fallback` pins it properly. I traced the discrimination rather than taking the claim: at divisor 2 the R2 GET's timeout clamps to `min(per_attempt, 1s)`, the retry loop declines a backoff that would overrun `r2_deadline`, and the RPC arm gets the remaining ~1s; at divisor 1 (or with the division removed) R2 owns all 2s and the fallback starts already expired, so `result.is_ok()` fails. The third assertion — `elapsed >= budget / R2_WITNESS_BUDGET_DIVISOR - 100ms` — is a positive control I didn't ask for and the better half of the test: it also catches the inverse failure where the R2 arm bails out early and the test passes without the budget cap ever mattering. Expressing it in terms of the divisor rather than a hardcoded 1s is the right call too — it pins the invariant, not the tuning constant.

**2 + 3. R2 misconfigurations (was blocking).** Better placement than I suggested. I said "one filter in the match arm"; hoisting both checks into `validate_args` puts them at the boundary with the rest of the startup validation, and the error text naming the empty-env-injection shape is what someone reading a crashed pod will actually need. Two things I checked specifically: the now-unreachable `warn!` arm is deleted rather than left as a dead branch, and `validate_args_rejects_r2_misconfigurations` opens with a positive control (`is_ok` on the full quad) before looping the four empty cases, so it can't pass as a validator that rejects everything.

**4. Queue wait.** You picked a third option better than either I offered. Keeping `witness_r2` end-to-end is the honest number on a request path, and splitting the queued share onto its own pre-registered `debug_trace_r2_witness_queue_wait_seconds` gives the cap-tuning signal without overloading the latency histogram — with the *why* recorded at both the metric and the adapter, which was the actual ask.

**5. Adapter dedup — and you're right, I was wrong about the target.** I proposed hoisting the shared error into `stateless-r2` without checking the dependency direction. Confirmed your point: `stateless-r2` has no `stateless-common` edge today, and it's consumed by mega-reth's uploaders, so hoisting there would drag the RpcClient stack into that build graph for a leaf crate's benefit. `stateless-common` as the host is the right direction. Happy for it to be its own PR.

One note for that follow-up, non-blocking and not worth another round here: the new queue-wait histogram only records on the success path — `get_block_object(...).await?` returns early on error, and `queue_wait` lives on `FetchedObject`, so it structurally can't exist on the failure side. That means `kind="deadline"`, the "gave up while queued" case and by definition the worst queueing, never reaches the histogram. `errors_total{kind="deadline"}` covers it qualitatively so the pair is adequate in practice; worth a look when the shared adapter is being reshaped anyway.

CI: `lint`, `test`, `no-std`, CodeQL and coverage all green. `pr-review` is red here and on every other open PR in the repo — the org `claude-pr-review` action dies with `ENOENT ... symlink` before reviewing — and it isn't one of the required checks.

Troublor (approved):

batatmosfer-byte:
Hello. 1. Initially, many actions were taken practically blindly due to a
lack of expertise. 2. The issue of conflicting actions by what is
essentially the same user (me) stems from having multiple accounts—two or
even four. The biggest problem is the long gap—about 12–15 years—between
creating the original, "correct" Google profile (the primary source) and
setting up a new Google portfolio. Attempts to integrate these
profiles—specifically the Gmail accounts—have been difficult; the goal is
to merge four or five email addresses into a single user identity. Google's
synchronization lags far behind modern technologies like GitHub or
Coinbase; having two different email addresses often causes verification
failures. Please provide guidance on how to fix the 404 error. Thank you
and best regards, Maciej Paweł Pawlak "TeRa4Yo2" ***@***.*** ✅
***@***.*** ✅ ***@***.***✅ ***@***.*** ✅
***@***.*** ✅ ***@***.*** ✅

pt., 7 sie 2026, 09:32 użytkownik liquan.mega ***@***.***>
napisał:

PR #176, opened 2026-08-06
perf(debug-trace-server): execute inbound JSON-RPC batch entries concurrently

flyq:
## Summary
jsonrpsee's built-in batch handling awaits entries strictly one after another (`jsonrpsee-server-0.26.0/src/middleware/rpc.rs:157`), so a batch answers in the sum of its items' latencies even when the server is otherwise idle — for batch clients the throughput floor is `batch size × per-item latency`, independent of server capacity. This PR adds an RPC middleware layer that executes the entries of one inbound batch concurrently (bounded), so a batch completes near its slowest entry.

## Motivation
Wave benching on tko (15 parallel JSON-RPC batches × 20 consecutive historical blocks per wave, the dominant external consumer's exact shape — the gateway forwards batches to the origin intact) measured per-item server time ~173ms but per-batch p50 = 3372ms ≈ 20 × 173ms across three waves, with only ~15 items in flight server-wide and every downstream limiter idle. Sequential batch execution is the throughput ceiling for that traffic (~70 items/s), and on deployments with higher per-item witness latency it pushes whole batches toward client timeouts.

## Implementation
- New `bin/debug-trace-server/src/rpc_middleware.rs`: `ConcurrentBatchLayer`, installed via jsonrpsee's `set_rpc_middleware`; it overrides only `RpcServiceT::batch` — single calls and notifications pass through untouched.
- Entries that failed parsing answer immediately; each remaining entry becomes one future over the inner service's `call`, driven by `buffer_unordered(N)` — the regular per-request pipeline (single-flight, caches, witness/R2 semaphores, metrics) applies per entry unchanged, and those semaphores keep providing downstream backpressure.
- Responses append in completion order; JSON-RPC 2.0 permits concurrent processing and order-independent responses (matched by `id`), and the fronting gateway already streams batch responses unordered.
- Tail semantics mirror the sequential implementation: the batch response is capped by the same `max_response_body_size` (over-limit collapses to the single too-big error), notification-only batches return an empty reply, empty batches an invalid request.
- `--batch-item-concurrency` / `DEBUG_TRACE_SERVER_BATCH_ITEM_CONCURRENCY` (default 16, min 1) bounds concurrently executing entries of one batch so a single huge batch cannot monopolize downstream permits against concurrently served requests; `1` restores strictly sequential execution and doubles as the kill switch.
- New `debug_trace_batch_size` histogram records entries per inbound batch.

## Testing
- `rpc_middleware::tests` run a real jsonrpsee server through the layer: a batch of 8 × 200ms methods completes in <800ms wall (sequential needs 1.6s; observed ~200ms), concurrency 1 sums latencies and preserves request order, a mixed batch (call + unknown method + malformed entry + notification) yields exactly the three per-entry responses, and an over-limit batch response collapses to a single error object.
- CLI test covers the default, CLI/env overrides, and rejection of `0`.
- `cargo test -p debug-trace-server`: 121 passed; workspace clippy/fmt/cargo-sort clean.

## Notes
- Expected effect on the wave bench: per-batch p50 ~3.4s → ~0.4s and wave throughput ~70 → 300+ items/s; will validate on the tko deployment with the same bench before rollout.
- Docs updated (README feature section, AGENTS.md); the bin now declares its tokio dev-dep features (`macros`/`rt-multi-thread`/`time`) explicitly instead of leaning on feature unification.


bin/debug-trace-server/src/rpc_middleware.rs
```diff
@@ -0,0 +1,347 @@
...
+//! batch answers in the sum of its items' latencies even when the server is otherwise
+//! idle — for batch clients the throughput floor is `batch size × per-item latency`,
+//! independent of server capacity. JSON-RPC 2.0 explicitly allows processing batch
+//! entries concurrently and responding in any order (responses are matched by `id`),
+//! so this layer replaces just the batch path: entries still go through the inner
+//! service's `call` one by one — every per-request mechanism (single-flight, caches,
+//! semaphores, metrics) applies unchanged — but up to `concurrency` of them run at
+//! once, and the batch completes near its slowest item. The bound exists so a single
+//! oversized batch cannot monopolize what its entries actually contend on — the
+//! witness-fetch semaphores and the runtime's worker threads (traces execute inline on
```
vincent-k2026: Non-blocking, but I'd fix this one here — it's the sentence someone will read at 3am while tuning the knob.

`buffer_unordered` multiplexes the entry futures inside a *single* task; it doesn't hand them to separate runtime workers. Since traces execute inline (no `spawn_blocking` in the path), the entries of one batch can only ever occupy **one** worker at a time for their CPU portions — so `--batch-item-concurrency 16` doesn't give a batch 16 cores, and the bound isn't what's keeping a batch off the worker threads. The witness-semaphore half of the rationale is right; the worker-thread half describes a monopolization that the single-task structure already prevents.

Same overstatement in the PR body's "the batch completes near its slowest entry" — true while entries are I/O-bound, which is the measured shape, but with the block-data cache warm an entry is close to pure CPU and the batch degenerates back toward sequential. Worth scoping both claims to the await-point overlap that the bench actually measured.

The honest fix for the CPU half is spawning entries as independent tasks, and I don't think that belongs here: `tokio::spawn` needs `'static`, the entries borrow from the request buffer, so it's `into_owned()` plumbing plus a fresh argument about what bounds worker occupancy — a follow-up PR, not a note in the margin of this one.

bin/debug-trace-server/src/rpc_middleware.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    use super::*;
+
+    const SLOW_MS: u64 = 200;
+
+    fn test_module() -> RpcModule<()> {
+        let mut module = RpcModule::new(());
+        module
+            .register_async_method("slow", |_, _, _| async {
+                tokio::time::sleep(Duration::from_millis(SLOW_MS)).await;
```
vincent-k2026, line 252: Non-blocking, coverage gap that follows from the note above: `slow` is a pure timer, so `batch_entries_execute_concurrently` proves the await-point overlap and nothing else. Every real trace could serialize on the synchronous EVM section and this test would stay green at exactly the same wall time.

That's not a vacuous test — remove the layer and it does go red (8 × 200ms against an 800ms bound) — it just doesn't reach the shape production trends toward as the block-data cache warms. A second method that burns CPU rather than sleeping (a spin to a deadline) would pin the real ceiling, and today it would document the sequential-CPU behavior rather than contradict it. Fine as a follow-up alongside the spawning question, since the two answers move together.

bin/debug-trace-server/src/rpc_middleware.rs
```diff
@@ -0,0 +1,347 @@
...
+    max_response_body_size: usize,
+    metrics: BatchMetrics,
+}
+
+impl ConcurrentBatchLayer {
+    /// `concurrency` must be at least 1 (the CLI parser enforces it);
+    /// `max_response_body_size` must match the server's configured limit so batch
+    /// responses are capped identically to jsonrpsee's own batch path.
+    pub(crate) fn new(concurrency: usize, max_response_body_size: usize) -> Self {
+        debug_assert!(concurrency >= 1, "a zero bound would stall every batch stream");
```
vincent-k2026: Non-blocking nit: `debug_assert!` compiles out in release, and `buffer_unordered(0)` doesn't fail loudly — it stalls the batch stream forever. Unreachable today (clap's `range(1..)` gates the only caller and the constructor is `pub(crate)`), so this is about which mechanism is doing the work: `Semaphore::new(n.max(1))` in the R2 fetcher makes the same guarantee structurally, and `concurrency.max(1)` here would too, for one character more than the assert.

vincent-k2026 (approved):
Reviewed against HEAD `7c50fc3`. Approving — the change does what its bench measured, the risky edges are covered, and `--batch-item-concurrency 1` is a genuine kill switch. Three non-blocking notes below; only the first is one I'd like fixed in this PR, and it's two sentences of prose.

The part I want to call out specifically is `tail_semantics_match_jsonrpsee`. Standing up a bare jsonrpsee server alongside the layered one and diffing the raw wire output for `[]` and the notifications-only shape turns "I mirrored upstream's tail semantics" — normally the most rot-prone claim in a middleware — into something that breaks loudly when an upstream bump moves the tail. That's the right instinct, and it's rarer than it should be. `concurrency_one_is_sequential_and_ordered` is the other one: a negative control on a concurrency change, asserting that the kill switch restores *both* summed latency and request order, so the escape hatch is known-good rather than assumed. The two comments explaining why `from_fn` and a named `async fn` are needed (rustc's higher-ranked lifetime inference inside `buffer_unordered`) are exactly the kind of why-comment that saves the next person an hour.

Notes 1 and 2 are the same underlying fact, and the open Codex thread on this PR got there first — its premise checks out, so it's worth resolving with a reply rather than leaving it hanging: there is no `spawn_blocking` or `block_in_place` anywhere in `rpc_service.rs` or `tracing_executor.rs`, and `tracing_executor::trace_block` is a plain synchronous `pub fn`, so EVM execution really does run inline in the entry future's poll.

None of this changes my read that the win is real: at ~173ms per item dominated by upstream witness/block/contract fetches, the await-point overlap is where the 3372ms → ~400ms comes from, and you've already said the number gets re-measured on tko before rollout.

bin/debug-trace-server/src/rpc_middleware.rs
flyq: The review raced the fix by a few hours: 258c236 implements exactly the spawning you describe as the honest fix — entries are converted to owned form lazily at spawn time (`Request::owned` + `Cow::into_owned` plumbing) and run through a bounded `JoinSet` refill-and-drain loop, with the sequential path's panic semantics preserved (`resume_unwind` on the connection task) and drop-aborts on early return/disconnect. The paragraph you're quoting was rewritten in that commit to describe the spawned structure, so the worker-thread half of the rationale is now literally true rather than aspirational. It landed in-PR rather than as a follow-up because a bot comment flagged the same single-task multiplexing the same day and the `'static` plumbing turned out small; the fresh argument you anticipate — what bounds worker occupancy once entries genuinely parallelize — is real and is ledgered as the follow-up (global trace-execution cap), which this change makes more pressing, not less.

bin/debug-trace-server/src/rpc_middleware.rs
flyq: Covered by 258c236: `cpu_bound_entries_parallelize` drives a batch of 6 through a worker-blocking method on a 4-worker runtime and asserts completion near 2 × the unit time — the single-task interleaving you're describing needs the full 6 × and goes red. One deliberate divergence from your sketch: a blocking sleep rather than a spin-to-deadline. It pins the same property — worker-thread occupancy — while staying independent of the CI host's core count (four sleeps overlap on four worker threads even on one core; four spins do not), so it can't flake on small runners. d52f838 records that rationale in the test doc so the choice doesn't read as an accident.

bin/debug-trace-server/src/rpc_middleware.rs
flyq: Agreed — fixed in d52f838: `.max(1)` in the constructor, assert dropped, mirroring the R2 fetcher's `Semaphore::new(n.max(1))`. Your argument is the stronger one, and it got stronger still with the spawned implementation: a zero bound's failure mode there is not even a stall but a non-empty batch silently answered as empty — exactly the kind of quiet wrongness a structural floor should make unrepresentable rather than a release-stripped assertion.

Troublor (approved):

PR #177, opened 2026-08-09
chore: bump version to 2.0.17

flyq:
## Summary

Production has been running #167 / #174 / #175 / #176 since 2026-08-07 while every binary still reported `2.0.16`, so during the fra/ore triage the version string could not be used to tell which code a host was on. `Cargo.lock` regenerated with `cargo update -w`.

## Testing

`cargo check` clean; no code change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1886,7 +1886,7 @@ dependencies = [
 
 [[package]]
 name = "debug-trace-server"
-version = "2.0.16"
+version = "2.0.17"
 dependencies = [
  "alloy-consensus",
  "alloy-evm",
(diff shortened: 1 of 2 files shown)


flyq:
Closing: no new version until the outstanding P0 work lands. Bumping now would stamp 2.0.17 on a build that still returns client-visible timeouts on the frontier witness path (see #179, which is itself not yet sufficient — the grace only covers half the witness budget and the fall-through can still stall on the public gateway).

The branch `liquan/chore/bump-2.0.17` is kept as-is and can be reopened once the P0 items are done; it is a one-line `Cargo.toml` change plus a regenerated `Cargo.lock`.

PR #184, opened 2026-08-10
chore: upgrade to reth v2.3.0 / revm 40 / alloy 2.0.5 (mega-evm revm-40 branch)

flyq:
## Summary

Switches the workspace from the reth v1.6.0 era to the reth v2.3.0 stack, following mega-evm's revm-40 upgrade (megaeth-labs/mega-evm#365): reth git tags v1.6.0 → v2.3.0, revm 27.1.0 → 40.0.3, revm-inspectors 0.27.3 → 0.40.1, alloy 1.0.23 → 2.0.5, alloy-evm 0.15.0 → 0.36.0, alloy-hardforks 0.2.7 → 0.4.7, alloy-primitives 1.3.0 → 1.6.0, alloy-trie 0.9.0 → 0.9.4, op-alloy 0.18.12 → 2.0.0, reqwest 0.12 → 0.13 across the workspace including `stateless-r2`'s explicit pin (its API exposes `&reqwest::Client` and must match mega-reth, whose `upstream/reth` workspace is on 0.13), leaving a single reqwest in the lockfile, and mega-evm tag v1.7.0 → branch `cz/chore/upgrade-revm-40` (locked at 8d1fc07). salt stays at v1.0.5. MSRV moves to 1.94.

## OP crates: one source identity

crates.io has no op-revm / alloy-op-evm release built against revm 40, so mega-evm pins the whole OP family to the Optimism monorepo at rev `f67d87cd`; this workspace must share that exact source or `OpTxEnvelope` / receipt / hardfork types split into incompatible copies. `alloy-op-evm`, `alloy-op-hardforks`, `op-alloy-consensus/network/rpc-types`, and `reth-optimism-chainspec` (which moved out of paradigmxyz/reth into the monorepo) are all pinned there, grouped under `# op` in the manifests; the monorepo's own reth deps use `tag = "v2.3.0"`, the same source as our direct reth pins, so the graph unifies with a single copy of every crate (verified in Cargo.lock).

## Withdrawal MPT verification reimplemented on alloy-trie

reth v2.3.0's sparse tries moved to a V2 node model that merges extension nodes into their child branches. An extension whose child branch is absent from the witness — a legal minimal exclusion proof the frozen MegaETH witness generator emits, whose split-on-insert never reads the child — cannot be represented: every reveal path silently drops it and the insert then fails on a blind node (caught by the synthetic fixture at block 360; both v2.3.0 trie impls also panic on standalone extensions, and reth's own flat-witness entry point shares the gap with no in-tree users). Upstream sidesteps this by making its internal proof producers over-retain extension children (`AddedRemovedKeys`), a contract our frozen witness format — and the immutable historical witnesses in R2 — can never satisfy.

`MptWitness::verify` (`crates/stateless-core/src/withdrawals.rs`) now linearizes the witnessed storage trie into `alloy_trie::HashBuilder` — the same primitive reth's `eth_getProof` generator is built on — as a sorted stream: `add_leaf` for surviving/updated leaves (leaf paths are the hashed slots and never change) and `add_branch` for unwitnessed subtrees (position and hash invariant under edits elsewhere). All structural changes — extension splits, branch collapses, new branches — fall out of the stream shape inside the builder, so the module contains no trie mutation code; this also returns to the storage-trie-only design of the original implementation (#16), dropping the synthesized account-leaf wrapper #119 introduced. A pre-pass with no updates must reproduce `storage_root` bit-for-bit before the post-pass output is trusted (skipped entirely when a block carries no withdrawal updates); updates the witness cannot prove — descent into an unwitnessed subtree, a collapse adopting an unwitnessed node — fail closed, matching the previous sparse-trie behavior. The `reth-trie-common` / `reth-trie-sparse` dependencies are gone. New tests cover the extension-split regression, collapse merging, unwitnessed-survivor and unwitnessed-descent fail-closed paths, and diverging-removal no-ops.

## Trace server: output drift and request gating

Struct-logger traces drift in serialization only (execution is byte-identical per mega-evm's replay gate): logs gain `"refund": 0`, memory words and `returnValue` gain `0x` prefixes — post-upgrade mega-reth serializes the prefix too, so the old strip-shim is removed. Gas values are unchanged (`tx_gas_used()` equals the old `gas_used()` with EIP-8037 pinned off). The new alloy builtin `erc7562Tracer` is not implemented: `RequestShape` classifies it as `Unsupported` and `classify_and_gate` rejects it with `-32602` before any block data is fetched, the executor arms remaining as unreachable backstops. Per-tx `EvmState` clones the revm-40 API made avoidable are dropped, and tracing inspectors are taken by `mem::replace` instead of cloning their recorded trace arenas. Since alloy 2.x, an unknown tracer name inside a mux config parses as a JS-tracer key and is rejected at `MuxInspector` construction rather than by serde; the error stays request-attributable with the same `tracerConfig` wording (#169's defense-in-depth test keeps its contract).

## Other API adaptations

`AccountInfo` gains `account_id: None` (witness state is address-keyed), `BlockEnv.slot_num` projects from `header.slot_number` like every other field (EIP-7843; absent until Amsterdam), `BundleState::state` uses the `AddressMap` fixed-bytes hasher, and mega-evm's `run_transaction` outcome carries `result_and_state` instead of separately movable fields.

## Testing

`cargo test --workspace`: 373 passed / 0 failed (the ignored suites are the pre-existing env-gated live-server tests), including the mainnet single-block fixtures, the synthetic-chain pipeline test that exposed the trie regression, and the new withdrawal edge-case tests. `cargo clippy --workspace --all-targets --all-features`, `cargo fmt --all --check`, and `cargo sort --check` are clean. Latest `main` (#169, #172) is merged in.

## Notes

mega-evm is pinned to a branch while megaeth-labs/mega-evm#365 is open; re-pin to a tag once it lands. mega-reth's `upstream/reth` branch pins the Optimism monorepo at `op-reth/v2.3.3` (2063657) rather than `f67d87cd` — if mega-reth ends up embedding stateless-core, the repos must converge on one monorepo rev or the OP types will not unify.


Cargo.toml
```diff
@@ -13,52 +13,55 @@ resolver = "2"
...
-reth-ethereum-forks = { git = "https://github.com/paradigmxyz/reth", tag = "v1.6.0", default-features = false }
-reth-optimism-chainspec = { git = "https://github.com/paradigmxyz/reth", tag = "v1.6.0", default-features = false }
-reth-trie-common = { git = "https://github.com/paradigmxyz/reth", tag = "v1.6.0", default-features = false }
-reth-trie-sparse = { git = "https://github.com/paradigmxyz/reth", tag = "v1.6.0", default-features = false }
+reth-ethereum-forks = { git = "https://github.com/paradigmxyz/reth", tag = "v2.3.0", default-features = false }
 
 # revm
-revm = { version = "27.1.0", default-features = false }
-revm-inspectors = { version = "0.27.3", features = ["std", "js-tracer"], default-features = false }
+revm = { version = "40.0.3", default-features = false }
```
Troublor, line 63: This bump breaks on-disk compatibility of the `CONTRACTS` table, and an in-place upgrade halts the node rather than degrading.

`Bytecode`'s serde layout changed between revm-bytecode 6.2.2 and 11.0.1: it went from a derive-serde enum ordered `Eip7702, LegacyAnalyzed` to a hand-written `BytecodeSerde` ordered `LegacyAnalyzed, Eip7702`. bincode encodes the variant by index, so every stored `LegacyAnalyzed` row (tag `0x01`) is re-read as `Eip7702 { delegated_address }` and fails the 20-byte `Address` length check. Compiling both versions side by side against `config::standard()`:

```
old encoded 38 bytes: [01, 09, 60, 80, ...]   # 6.2.2  Eip7702=0, LegacyAnalyzed=1
NEW DECODE ERR: OtherString("could not convert slice to array")
new encoded 38 bytes: [00, 09, 60, 80, ...]   # 11.0.1 LegacyAnalyzed=0, Eip7702=1
```

There is no schema-version key in `stateless-db` and no `keccak256(code) == hash` re-check on the read path, so `StoreError::Corrupt` propagates untouched through `ContractCache::get` (`cache.rs:144`) into the pipeline, whose default classification is fatal (`pipeline/traits.rs:82-83`). A node that keeps its `--data-dir` — the documented normal operation (`README.md:73`: `--genesis-file` is "required on first run, stored in database for subsequent runs") — dies on the first block touching any previously cached contract. Rollback breaks symmetrically. There is also a rare silent branch: a stored contract whose bytecode is exactly 20 bytes decodes "successfully" into a delegation to a garbage address.

The cheapest correct fix makes the cache self-validating, which also covers every future encoding change — in `read_contracts` (`crates/stateless-db/src/helpers.rs:189`), treat a decode failure or a `keccak256(code) != hash` mismatch as a miss rather than an error, so a stale row is simply re-fetched. The alternative — a schema-version key checked on DB open — is more explicit but needs a migration path for each table. Either way this needs an operator-facing note that the data dir must be cleared for this release.

Verification: a test that writes a fixture blob with the 6.2.2 layout (`0x01` variant tag) and asserts the read path returns a miss, not an error.

`WITNESSES` and `BLOCK_DATA` are not affected: `LightWitness` is unchanged and `salt` stays at v1.0.5, and stored blocks are JSON whose new `blockTimestamp` is `#[serde(default)]` on both the alloy and op-alloy sides.

_(verified failure scenario)_

flyq: fixed in [2f90a7b](https://github.com/megaeth-labs/stateless-validator/pull/184/commits/2f90a7b3ed7d14f81e2c55d9e35619f95c4990fd)

vincent-k2026:
## Blocking

**1. `mega-evm` is pinned to a branch.** The lockfile holds today's rev, but any `cargo update` or force-push silently swaps the consensus execution engine, and every other pin in this manifest is a tag (`salt` v1.0.5, `reth` v2.3.0). Needs megaeth-labs/mega-evm#365 merged and tagged first.

**2. Withdrawals MPT verification is a new bespoke `Linearizer` (+433/−99), covered by 8 hand-built trie shapes and 31 mainnet fixtures that all sit below block 15,412,411.** The algorithm itself reads correct: emission order is monotonic across the leaf / extension / branch arms, `split_at_prefix` holds, the sole-survivor-is-unwitnessed collapse guard fails closed, and the no-update pre-pass self-checks the same walk on every witness. But consensus-critical code with no randomized coverage shouldn't merge on hand-picked shapes.

Before merge: a differential test. Random storage trie → reference root from `HashBuilder` over the full state → prune to a minimal witness → random insert / overwrite / remove / remove-absent set → assert `compute_root` either matches the reference or returns `TrieOperationFailed`. Never a plausible-but-wrong root.

Three smaller ones in the same file:

- `child_regions` ends on `debug_assert!(rest.is_empty())` (`withdrawals.rs:373`), which disappears in release. The invariant it guards is alloy-trie's iteration semantics: `BranchChildrenIter` walks `CHILD_INDEX_RANGE` (0..16) and yields `(i, None)` for unset slots, so the `idx == 15` round drains `rest`. That is also why Codex's P1 on this line doesn't hold — an update targeting an empty slot lands in a `child: None` region and streams through `add_insert_leaves`. But if that iteration ever narrows to populated children, release builds silently drop the update and reject a valid block. Make it a hard error, and add the case (insert into an empty slot whose nibble is above the last populated child) to the tests.
- `compute_root` returns `Ok(EMPTY_ROOT_HASH)` when the root node is absent. Fail-open in shape — safe only because the caller always runs the pre-state pass first. `Err(trie_error("witness lacks the root node"))` says what happened and doesn't owe a debt to call order.
- The pre-state pass runs on unauthenticated witness bytes, and `HashBuilder::add_branch` is an `assert!` on key ordering (alloy-trie 0.9.5, `src/hash_builder/mod.rs:164`). Unreachable for any decodable witness if the emitted keys are structurally monotonic — that argument belongs in `verify`'s doc. Separately, `emit_subtree` recurses and an empty-key extension consumes no depth, so a large enough malformed witness could blow the stack; fine to just note it if witness size is bounded upstream.

**3. Conflicting, and the conflict has semantics.** main has since landed the R2 custom-domain / h2 / `--r2-connections` work that `crates/stateless-r2/Cargo.toml` here still says to "revisit alongside". The rebase has to re-settle reqwest 0.13's renamed features (`rustls-tls` → `rustls`), `http2` now being opt-in, and whether `stateless-common`'s `http1_only` pin still holds under it. Full suite needs re-running afterwards; the CI on this branch predates all of it.

## Needs evidence

**4.** Dropping the `0x`-prefix strip shim on struct-logger `returnValue` is a consumer-visible wire change: if production mega-reth is still on alloy 1.x, the trace server diverges from the node it mirrors. Paste a same-tx raw-JSON comparison. The return type also moved from `GethTrace::JS(frame_value)` to typed `GethTrace::from(DefaultFrame)`, so the diff surface also covers `"refund": 0` and memory-word prefixes — same run.

**5.** `result.gas_used()` → `result.tx_gas_used()` lands in callTracer / struct-logger `gasUsed`. "Equal with EIP-8037 pinned off" needs the two revm 40 definitions or a measured diff.

**6.** Preceding txs now replay under `TracingInspectorConfig::none()`. Hot-path change riding a perf cleanup — one "same block, same tx, byte-identical trace before/after" run closes it.

## Non-blocking

- `write_add_contracts` accepts a `(hash, bytecode)` pair whose hash doesn't match. The read side catches it now, but the invariant belongs at the write site: `debug_assert_eq!(bytecode.hash_slow(), *hash)`.
- `hash_slow()` on every disk hit is a per-block path; worth a number.
- `rust-version` 1.95 → 1.94 in `AGENTS.md` reads like an MSRV regression until you notice Cargo.toml went 1.91 → 1.94. One line in the body would help.

PR #187, opened 2026-08-14
feat(stateless-r2): custom-domain R2 target with HTTP/2, keeping the raw S3 endpoint path

flyq:
## Summary
`stateless-r2` gains a second read target: a **Cloudflare custom domain** fronting the witness bucket — unsigned GETs of `/{key}` (no bucket segment), optional Cloudflare Access service-token headers (`CF-Access-Client-Id/Secret`), and HTTP/2 via the reqwest `http2` feature with adaptive flow-control windows and keep-alive pings. `--r2-connections` (default 1, so nothing changes unless it is set) spreads those GETs over several HTTP/2 connections. The SigV4 S3 path is byte-for-byte unchanged and now pinned `http1_only`, so its wire behavior is independent of the feature flip. Both binaries expose the target as `--r2-custom-domain` plus the optional `--r2-access-client-id`/`--r2-access-client-secret` pair (the id is redacted like the secret — it alone can look up the token); the two targets are mutually exclusive, enforced post-parse so the error names both flags even when they arrive through env vars.

## Why
R2's bare S3 endpoint negotiates HTTP/1.1 only (ALPN-verified live), so N in-flight GETs hold N connections, capped by the egress IP's adaptive connection budget — the constraint behind today's `--r2-max-concurrent-requests=48`. A custom domain rides the h2-capable CDN edge: concurrency decouples from connection count, the steady-state handshake load collapses to a few long-lived multiplexed connections instead of a socket per in-flight GET, and the immutable witness objects become edge-cacheable. *Steady state* is the qualifier that matters: version selection is pure ALPN, so hyper takes no connect lock before the handshake and every queued GET opens its own until one lands in the pool. Cold start and every reconnect therefore still burst up to the per-connection concurrency — the mechanism behind the `kind=connect` failures reported below.

## Where one connection stops, and why `--r2-connections` exists
One `reqwest::Client` is one HTTP/2 connection, and hyper opens no second one to relieve a saturated one. The pool hands every checkout the same connection back — an h2 entry is reserved as a shared clone and re-inserted on each take (`Reservation::Shared`, `IdlePopper::pop`) — and the only health check it applies is `PoolClient::is_open`, which reports liveness rather than stream capacity: it delegates to `hyper::client::dispatch::UnboundedSender::is_ready`, which is `!giver.is_canceled()`, over a dispatch channel that is unbounded. A live-but-saturated connection is therefore always found and reused, and the connect path is never entered on its account.

(`Pool::connecting`, which does serialize h2 connects, is not what does this here: its lock is gated on `ver == Ver::Http2`, which hyper-util sets only under `http2_only(true)` — reqwest's `http2_prior_knowledge()`. This target deliberately selects by ALPN instead, so that the plaintext loopback path the mocks use keeps working, which leaves `ver == Ver::Auto`.)

So a request past the edge's `SETTINGS_MAX_CONCURRENT_STREAMS` neither fails nor earns a connection of its own — it waits inside the connection, where the wait still counts against the per-attempt timeout and escapes the permit-queue accounting: `debug_trace_r2_witness_queue_wait_seconds` never sees it, and on the validator, which publishes no queue-wait series and subtracts permit wait out of `witness_fetch_r2_time_seconds`, it lands in that histogram as R2 time.

That makes the edge's per-connection stream limit a **per-process ceiling**, and `--r2-connections` is the only thing that lifts it. Connections are handed out round-robin **per attempt** rather than per fetch, so a retry leaves the connection that just failed instead of retrying into the same fault. The in-flight cap still bounds all of them together, split evenly and rounded up — rounding down would leave a connection at zero permits and wedge every GET routed to it — so raising the connection count alone spreads the same concurrency thinner rather than lifting the ceiling. That cap is `--r2-max-concurrent-requests` on the trace server; on the validator it is `--witness-max-concurrent-requests`, which sizes the RPC gateway too, so there the two consumers trade off against each other.

The availability argument is what justifies landing this alongside the target rather than after it: a single connection is a single point of failure, and when it drops every GET riding it fails together. The validator's R2 mode has no RPC fallback to absorb that.

## Acceptance measurements

Measured on one host against the delivered domain. The probe drives the production fetcher through a `path` dependency; it lives with the ops tooling and is not part of this diff.

**The stream limit is read, not assumed.** Re-read it per zone — the whole concurrency story hangs on this one number.

```
$ nghttp -nv https://<domain>/
The negotiated protocol: h2
[recv] SETTINGS frame
  [SETTINGS_MAX_CONCURRENT_STREAMS(0x03):100]
```

### What this target is worth, on the workload it was built for

The same host ran a one-hour soak against the **S3** target on 2026-08-11 in a fixed shape — 15 concurrent JSON-RPC batches of 20 blocks, 300 blocks in flight, `debug_traceBlockByNumber` + `callTracer` over historical blocks. Re-running that exact shape against the custom domain, on this branch:

| | S3 endpoint, 2026-08-11 | custom domain, this branch |
| --- | ---: | ---: |
| in-flight R2 cap | 48 | 1024 over 16 connections |
| **R2 semaphore queue wait** | **337 ms — 66% of the witness stage** | **0 ms** |
| witness stage | 509 ms | 83 ms |
| throughput | 195 blocks/s | **435 blocks/s** |

The old run's own conclusion named this fix: the S3 endpoint speaks HTTP/1.1 only, one request per connection, so 48 permits were the ceiling and two thirds of every witness fetch was spent waiting for one. The queue is now gone, and throughput on the unchanged workload is 2.2×.

Pushed further on the same build — five 300-second rungs from 300 to 1500 blocks in flight, each on a block range no earlier rung had touched — the target absorbed **1,175,400 blocks with zero R2 failures of any kind**: no `missing`, no `throttled`, no `status`, no `connect`, one `transport` in 1.6M GETs. R2 semaphore queue wait stayed at 0 ms throughout, and `witness_r2` served 100% of fetches with the generator and public gateway untouched on the request path. Witness stage grew only 83 → 197 ms across a 5× concurrency increase; what did grow was the local node behind `eth_getHeaderByHash`/`eth_getBlockByHash`, 2.4 → 115 ms and 3.9 → 125 ms. **R2 is no longer the constraint on this path** — which is the result this target was added to produce.

### Cold vs warm, and the negative-cache gate

Two single-pass runs over disjoint historical ranges, 32 connections × 100:

| | cold, 15.0M–20.0M | warm, 18.0M–20.0M |
| --- | ---: | ---: |
| objects | 4,927,355 | 1,950,494 |
| throughput | 17,919 obj/s | 28,507 obj/s |
| mean latency | 164 ms | 86 ms |
| `cf-cache-status` | MISS 2499 / 2500 | HIT 1000 / 1000 |

Like for like on the same blocks: **+59% throughput, −46% latency**. The gain is in the body of the distribution, not the tail — p50 and p90 each drop a bucket, the p99 bucket is unchanged, and the share over 1600 ms rises from 0.002% to 0.043%. Functional acceptance passed alongside: Access enforcing, credentials accepted, an authenticated fetch of a real object, ALPN `h2`, and a 404 that goes `MISS → EXPIRED` and never `HIT`, which is the negative-cache gate the deployment note below depends on.

### Failures, and what is not claimed

At 32 connections × 100 — one connection per 100 streams, exactly the edge's advertised limit — 1.40% of GETs failed cold and 2.43% warm, essentially all `kind=connect`; `transport` never exceeded 2 per minute. Stating only what follows from the code: version selection is ALPN, so hyper takes no connect lock before the handshake and every queued GET on a connection that has none yet opens its own, which makes a cold or dropped connection a burst of up to the per-connection share of simultaneous handshakes against the connect timeout. Some of the count is fan-out instead — the task that wins the post-ALPN lock leaves the rest queued, and `PoolInner::connected` clears that list when it fails. Which effect dominates is not established, and neither is why the connections drop. `..._r2_witness_errors_total{kind="connect"}` is the series to watch. None of this appeared at the concurrencies the trace-server workload actually uses: the five-rung ladder above ran clean.
## Deployment note
Any edge cache rule that makes these objects cacheable **must set 404s to bypass cache**: readers probe objects before the uploader PUTs them, and a negatively-cached 404 would pin a pre-upload frontier miss for the TTL (the validator's R2 mode has no RPC fallback) and can false-fire the below-band `kind="missing"` bucket-integrity alarm. The caveat is carried in the flag docs, README, and AGENTS.md; the ops runbook for the domain setup includes the concrete Cache Rule setting.

Sizing: what has to stay at or below the edge's stream limit is the **per-connection share** — the in-flight cap divided by `--r2-connections` — not the total, and the fetcher warns at startup when that share exceeds it. Raising `--r2-connections` on its own therefore buys fault isolation at unchanged throughput, which makes it a clean single-variable first step; raising throughput needs the cap raised with it. The count is published as `debug_trace_r2_connections` / `r2_connections`.

## Testing
Wire-shape tests through a new capturing mock (`mock_r2_capturing`): bucketless unsigned custom-domain path, Access headers present including on retry attempts, the S3 path's bucket prefix + SigV4 authorization pinned, bare 404 → `Missing`, redirects surfaced as `Status` (Access login bounce) on both clients, and `Debug` redaction of the Access pair through the fetcher. Sharding adds the per-connection permit split including the round-up that keeps a connection off zero, round-robin hand-out across connections, the saturated pool spreading evenly over every connection rather than a subset, and the startup advisory firing only on per-connection over-subscription — silent exactly at the limit, which is the intended sizing, and silent when unlimited, which is the default.

Every R2 argument rule is covered by `validate_r2_flags`'s own unit tests (`crates/stateless-common/src/r2_args.rs`), each asserting the error names the offending flag. The trace server's CLI tests drive those rules through `validate_args`: mutual exclusion naming both flags, the all-or-nothing Access pair, empty-value rejection (a blank `R2_CUSTOM_DOMAIN=` line over a working S3 config gets the named empty-value error, not a phantom target conflict), the shared tuning flags accepted with either target and rejected by name with none, a zero connection count, and a count set alongside the S3 endpoint. The validator's CLI test pins the other half — that all of those shapes parse — since its R2 flags are validated post-parse and only under `--witness-source r2`.

An IPv6 loopback origin is pinned end to end (`http://[::1]:8080` accepted with credentials, brackets surviving the origin rebuild, a public IPv6 host still refused over plaintext). `cargo test --workspace` (55 stateless-r2 + 55 stateless-common + 159 debug-trace-server + 40 stateless-validator), `cargo clippy --workspace --all-targets --all-features`, `cargo fmt --check`, and `cargo sort --check` are all clean.

## Notes for reviewers
The reqwest `http2` feature propagates by Cargo feature unification — first inside this workspace, then to consuming ones (mega-reth picks it up on the coordinated bump). Unpinned clients start *offering* h2 via ALPN, which servers are free to decline. Two clients here must not move, and this PR pins both `http1_only`: the S3 fetcher, which keeps the signed path's wire behavior fixed regardless of what the endpoint ever negotiates, and `stateless-common`'s shared JSON-RPC client (`rpc_client.rs`), whose endpoints do accept h2 — that would put the multi-MB witness payloads on one shared connection per host with adaptive windows off. Moving the RPC path to h2 is a change worth measuring on its own, not one to inherit from a feature flip.

Multi-connection sharding was originally deferred here pending measurements against the real domain. The measurements arrived and it landed in this PR instead: the availability argument turned out to be the stronger one, and it is not something a follow-up should leave open on a target whose validator mode has no fallback. Still out of scope: HTTP/3 (the edge advertises `alt-svc: h3`, but reqwest's h3 support is unstable and gated behind a `RUSTFLAGS` cfg), and recalibrating the deployed concurrency cap, which is a configuration change rather than a code one.


crates/stateless-r2/src/fetch.rs
```diff
@@ -229,34 +399,96 @@ impl R2ObjectFetcher {
...
+            return Err("Refusing to send Cloudflare Access credentials over plaintext http:// \
+                 to a non-loopback host: give --r2-custom-domain an https:// origin"
+                .to_string());
+        }
+        let mut builder = base_client(timeouts)
+            // Cloudflare's Browser Integrity Check, on by default on many zones, challenges
+            // requests that carry no user agent — which would arrive here as a non-retryable
+            // 403 on every GET. It also makes this traffic attributable in zone analytics.
+            .user_agent(concat!("stateless-r2/", env!("CARGO_PKG_VERSION")))
+            .http2_adaptive_window(true)
```
Troublor: Nothing here observes the protocol that was actually negotiated, so a domain that doesn't offer h2 via ALPN degrades to HTTP/1.1 completely silently — and the degraded state is worse than the S3 target it replaced.

There's no `http2_prior_knowledge()` (correctly — it would break the plaintext-loopback mock and port-forward path this code deliberately supports at `is_loopback_host`), so version selection is pure ALPN and an h1-only peer just returns an ordinary successful response. The three h2 knobs go inert, but `pool_idle_timeout(None)` on line 475 stays active and now governs an **HTTP/1.1** pool. Reqwest's defaults are `pool_idle_timeout: Some(90s)` and `pool_max_idle_per_host: usize::MAX`, and nothing in the workspace sets the latter — so idle sockets are never reaped at all, where the S3 client reaps at 90s. Meanwhile `r2_target_info{target="custom_domain"}` keeps asserting the intended target.

The triggers are all realistic: a grey-clouded DNS record, a non-Cloudflare origin, a zone with HTTP/2 disabled, or any plaintext `http://` origin (no ALPN at all). And since every custom-domain test in this PR runs against the plaintext HTTP/1.1 mock, the currently-green tests *are* the degraded path — which is exactly why nothing caught it.

Suggested fix, in two independent parts:

1. Observe `response.version()` on the `CustomDomain` arm and surface a mismatch. Warn **once** (an `AtomicBool`/`Once` on the fetcher) — a per-GET WARN would fire thousands of times a minute. Better still, extend the existing `r2_target_info` gauge from `{target}` to `{target, http_version}`: that gauge already exists to attribute the target-less R2 series during a rollout, and "did we actually get h2" is the same question, visible fleet-wide on a dashboard rather than in one host's log.
2. Add a finite `pool_max_idle_per_host` to this builder so the h1 fallback can't accumulate unreaped sockets. Cheap and worth doing regardless of (1).

A test is straightforward with what's already here: the existing `mock_r2` replies `HTTP/1.1`, so asserting the degradation signal fires against it locks in that today's passing tests are recognized as the degraded case.

*(verified failure scenario)*

bin/stateless-validator/src/app.rs
```diff
@@ -266,6 +296,24 @@ pub async fn run() -> Result<()> {
     // In R2 mode the RpcClient's witness providers are never used, but its constructor requires
     // a non-empty list — hand it the data endpoints as a placeholder.
     let data_apis: Vec<&str> = args.rpc_endpoint.iter().map(String::as_str).collect();
+    reject_empty_r2_values(&[
```
Troublor: These two checks run **before** the `match args.witness_source` on line 317, so they now apply to every startup — including `--witness-source rpc`, the default, where none of these seven values is ever read.

Before this PR the only R2 argument handling was the `require_r2` calls *inside* the `WitnessSource::R2` arm, so an RPC-mode validator ignored `--r2-*` entirely. That's a behavior change with no test pinning it either way.

The shape that bites: a fleet templating one env file across roles, where an unset variable renders as a blank line — `STATELESS_VALIDATOR_R2_SECRET_ACCESS_KEY=`. Clap treats an empty string as *present*, so `reject_empty_r2_values` fires and the process exits with `--r2-secret-access-key is set but empty`. Every RPC-mode validator in that fleet crash-loops on the next deploy — hosts this PR wasn't meant to touch, failing on config that's inert for them. `reject_conflicting_r2_targets` has the same reach.

To be explicit about what I can and can't back: the ordering is verified, the blank-env-template deployment shape is an inference. If every role gets its own explicit config file this is a non-issue and the strictness is arguably a feature — but then it's worth a comment saying the reach is deliberate. If it isn't deliberate, scoping is one move: put both calls inside the `WitnessSource::R2` arm, plus a test asserting `--witness-source rpc` starts cleanly with an empty `--r2-bucket` present.

Worth noting the trace server isn't affected the same way — its empty-value check was already unconditional in `validate_args` before this PR.

bin/stateless-validator/src/app.rs
```diff
@@ -403,14 +436,128 @@ fn override_ms(ms: Option<u64>, default: Duration) -> Duration {
...
+    Ok(client)
+}
+
+/// Rejects any `--r2-*` flag that is present but empty — the shape of a failed env injection.
+///
+/// Runs over every R2 flag before target selection, matching the trace server's order, so an
+/// empty value is diagnosed as itself rather than read as a configured target or as a leftover
+/// from one. Without this pass a blank `..._R2_CUSTOM_DOMAIN=` beside a working S3 config gets
+/// told to unset the S3 config.
+fn reject_empty_r2_values(values: &[(&str, Option<&str>)]) -> Result<()> {
```
Troublor: The R2 target-selection policy now exists twice: as free functions here (`reject_empty_r2_values` / `reject_conflicting_r2_targets`) and inline in the trace server's `validate_args` (`bin/debug-trace-server/src/main.rs:509-570`), with near-identical error strings and the same stringly-typed `(&str, bool)` flag arrays on both sides.

The evidence that this needs one home is in the PR itself — the copies have already drifted, three ways:

- the trace server names the *missing* members of an incomplete S3 quad (`main.rs:558-568`); this side has no equivalent, so a partial quad surfaces later via `require_r2` one flag at a time
- this side sweeps emptiness twice (`reject_empty_r2_values` up front, then `optional_r2`'s `Some("")` arm — consequently unreachable in production, though `optional_r2_distinguishes_absent_from_empty` still tests it); the trace server sweeps once
- the shared tuning flags are gated on the trace server (`requires = "r2_target"`, with a test for the orphan case) but **not** on the validator — `--r2-connect-timeout-ms` at line 150 has no `requires`, so passing it with no R2 target configured is silently accepted and ignored

Since these rules decide which `Target` the fetcher gets built with, they'd sit naturally in `stateless-r2`:

```rust
// in stateless_r2::fetch
pub enum TargetChoice<'a> {
    S3 { endpoint: &'a str, bucket: &'a str, access_key_id: &'a str, secret: &'a str },
    CustomDomain { domain: &'a str, access: Option<CfAccessCredentials> },
}

/// Resolves the operator's R2 flags into exactly one target, or the named error explaining
/// why not. `flag_names` lets each binary report its own flag spellings.
pub fn resolve_target<'a>(raw: &'a R2Args, flag_names: &R2FlagNames) -> Result<TargetChoice<'a>, String>;
```

Both binaries would then call one implementation with their own flag names, so a rule added on one side can't silently miss the other. Not blocking — but worth doing before a third consumer picks this up.

I considered just keeping the two copies in sync by convention and rejected it: they diverged three ways within a single PR, which is about as strong a signal as that convention ever gets.

bin/stateless-validator/src/app.rs
```diff
@@ -96,27 +97,56 @@ pub struct CommandLineArgs {
...
+    /// race, so **any edge cache rule making these objects cacheable must set 404s to bypass
+    /// cache** — an edge-cached 404 would otherwise pin every pre-upload frontier miss for the
+    /// negative-cache TTL and stall tip-following for minutes at a time.
+    #[clap(long, env = "STATELESS_VALIDATOR_R2_CUSTOM_DOMAIN")]
+    pub r2_custom_domain: Option<String>,
+
+    /// Cloudflare Access service-token client id, sent as `CF-Access-Client-Id` on every
+    /// custom-domain GET. Omit when the domain is locked by an IP allowlist instead.
+    /// Redacted like the secret: the id alone is enough to look up the token.
+    #[clap(long, env = "STATELESS_VALIDATOR_R2_ACCESS_CLIENT_ID", requires_all = ["r2_custom_domain", "r2_access_client_secret"])]
```
Troublor: The Access pair is the one R2 coupling rule left on clap, and it produces exactly the error this PR set out to eliminate. Built from head:

```
$ stateless-validator ... --r2-custom-domain https://w.example.com --r2-access-client-id tok
error: one or more required arguments were not provided

$ stateless-validator ... --r2-custom-domain https://w.example.com --r2-endpoint https://a.r2...
Error: --r2-endpoint and --r2-custom-domain are mutually exclusive R2 targets: configure exactly one
```

The second is the standard this PR sets, and the reasoning in the `Args` comment on the trace server says why: no `error-context` feature, so clap names no argument, which is useless to an operator whose only clue is an env file. That reasoning applies verbatim here. (It bites harder than usual because this workspace also builds clap without `help` — `--help` returns `error: unexpected argument found` — so these doc comments are source-only and the error message is genuinely all the operator gets.)

There's a second reason to move it. Both call sites build the credentials with `Option::zip`:

```rust
let access = optional_r2(&args.r2_access_client_id, "--r2-access-client-id")?
    .zip(optional_r2(&args.r2_access_client_secret, "--r2-access-client-secret")?)
```

`zip` yields `Some` only when both halves are; one half alone silently becomes `None`, which is a *legitimate* configuration (IP-allowlisted domain), so nothing downstream objects — you get a cleanly-constructed, working, **unauthenticated** client. Today that's unreachable because `requires_all` rejects it first, but the only thing between a half-configured token and a silent auth bypass is a clap attribute. If it's ever loosened, the edge 403s every GET as non-retryable `Status`, and validator R2 mode has no RPC fallback.

One change fixes both: enforce the pair in the same post-parse pass as the target rules, erroring by name on a half-set pair, and let the `zip` below rest on an invariant that's actually checked. Same applies to `main.rs:347/353`.

Troublor (changes requested):

bin/stateless-validator/src/app.rs
```diff
@@ -266,6 +291,7 @@ pub async fn run() -> Result<()> {
     // In R2 mode the RpcClient's witness providers are never used, but its constructor requires
     // a non-empty list — hand it the data endpoints as a placeholder.
     let data_apis: Vec<&str> = args.rpc_endpoint.iter().map(String::as_str).collect();
+    reject_dual_r2_targets(&args.r2_endpoint, &args.r2_custom_domain)?;
```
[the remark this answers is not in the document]
flyq: Addressed. R2 flag validation now runs only under `--witness-source r2`: `validate_r2_flags` is called from `build_r2_client` (`bin/stateless-validator/src/app.rs:461`), whose sole call site is inside the `WitnessSource::R2` arm. The `WitnessSource::Rpc` arm reads no `--r2-*` field, so a default-RPC startup with stale R2 env vars no longer aborts, and nothing rejects them at parse time either — the R2 flags carry no clap `conflicts_with`/`requires`.

The same reach applied to a blank env value on a numeric flag, which clap rejects before `run` can decide the flags are irrelevant. `--r2-connections` therefore takes text and is parsed post-parse (`parse_r2_connections`); `rpc_mode_tolerates_blank_and_conflicting_r2_values` in `tests/integration.rs` pins that a blank `--r2-bucket` / `--r2-custom-domain` / `--r2-connections` / `--r2-access-client-id`, and two conflicting targets, all parse in rpc mode.

bin/stateless-validator/src/app.rs
```diff
@@ -96,27 +96,52 @@ pub struct CommandLineArgs {
...
+    /// race, so **any edge cache rule making these objects cacheable must set 404s to bypass
+    /// cache** — an edge-cached 404 would otherwise pin every pre-upload frontier miss for the
+    /// negative-cache TTL and stall tip-following for minutes at a time.
+    #[clap(long, env = "STATELESS_VALIDATOR_R2_CUSTOM_DOMAIN")]
+    pub r2_custom_domain: Option<String>,
+
+    /// Cloudflare Access service-token client id, sent as `CF-Access-Client-Id` on every
+    /// custom-domain GET. Omit when the domain is locked by an IP allowlist instead.
+    /// Redacted like the secret: the id alone is enough to look up the token.
+    #[clap(long, env = "STATELESS_VALIDATOR_R2_ACCESS_CLIENT_ID", requires_all = ["r2_custom_domain", "r2_access_client_secret"])]
```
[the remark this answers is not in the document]
flyq: Addressed. The clap-level pairing is gone: `--r2-access-client-id` and `--r2-access-client-secret` carry only `long` and `env` (`bin/stateless-validator/src/app.rs:120`, `:125`), so a stale env var no longer fails parsing. The rule moved post-parse into `validate_access_pair` (`crates/stateless-common/src/r2_args.rs`), reachable only through `validate_r2_flags` in `build_r2_client`, which `run()` calls only from the `WitnessSource::R2` arm — every `--r2-*` flag is inert in RPC mode. `tests/integration.rs` pins that a half-set pair, and a full pair with no domain, both parse; the messages are covered by `a_half_set_access_pair_is_named` in `r2_args.rs`.

crates/stateless-r2/src/fetch.rs
flyq: Both parts are in, and the second one caught a further defect.

(1) The fetcher latches the first response's version on the `CustomDomain` arm (`crates/stateless-r2/src/fetch.rs`, `observe_version`), warns once naming what it actually negotiated, and exposes it through `negotiated_http_version()`; both binaries publish it as a `version`-labeled info gauge (`..._r2_negotiated_http_version_info`). `custom_domain_reports_the_protocol_it_actually_negotiated` asserts `Some("http/1.1")` against the plaintext mock, so the suite now records that these tests have been the degraded path all along — your point, pinned.

(2) `pool_max_idle_per_host` is finite, derived from the concurrency cap and defaulting to 64 when unset, so `pool_idle_timeout(None)` can no longer let an h1.1 fallback accumulate unreaped sockets.

The nuance you'd expect from (1) — that the gauge only appears once a fetch returns — turned out to be worse than that until just now: both adapters published only on the *success* path, so a zone answering 403 on every GET surfaced the downgrade through the WARN alone. Codex found the same thing independently from the other side. Fixed in 2f481c5: both adapters publish on any outcome, which is the case where the answer actually matters.

bin/stateless-validator/src/app.rs
flyq: Scoped, and the residual you'd have predicted was real.

Structurally: the empty-value and target-exclusion checks are now one call — `validate_r2_flags` at `bin/stateless-validator/src/app.rs:461`, reached only through `build_r2_client`, whose only call site is inside the `WitnessSource::R2` arm. In rpc mode no `--r2-*` string is read. The reach is deliberate on the trace server (no `--witness-source` there, and `R2Target::None` is a valid outcome) and scoped on the validator, which is what you asked for.

You flagged the blank-env-template shape specifically, and this branch had reintroduced exactly it: `--r2-connections`, added after your comment, was a `usize`, so `STATELESS_VALIDATOR_R2_CONNECTIONS=` aborted an rpc-mode startup at clap parse with `error: invalid value for one of the arguments` — no flag named, since we build clap without `error-context`. It now takes text and is parsed by `parse_r2_connections` where the other R2 rules run (2f481c5).

The test is in: `rpc_mode_tolerates_blank_and_conflicting_r2_values` (`tests/integration.rs`) pins that a blank `--r2-bucket`, `--r2-custom-domain`, `--r2-connections` or `--r2-access-client-id`, and two conflicting targets, all parse under `--witness-source rpc`. Being honest about its reach: it pins the parse layer, which is the half that just regressed. The gating itself lives in `run`'s dispatch and would need that dispatch extracted to pin end to end; for now it rests on the comment at `app.rs`.

bin/stateless-validator/src/app.rs
flyq: Single-sourced. The policy is now `stateless_common::validate_r2_flags` (`crates/stateless-common/src/r2_args.rs`), called by both binaries with their own flag spellings — validator via `r2_flags` in `app.rs`, trace server via `r2_flags` in `main.rs`. No R2 policy free functions remain in `app.rs`; `reject_empty_r2_values` and `reject_conflicting_r2_targets` are gone. Your three drifts: (a) the whole missing quad is named in one message, (b) emptiness is swept once, before target selection, and (c) the tuning flags now go through the same rule for both binaries.

On (c) the validator's answer is deliberately different, and the doc that contradicted it was the actual bug. It passes `tuning: &[]`, because under `--witness-source rpc` every `--r2-*` flag is inert by design — the scoping you asked for in the sibling thread — and under `--witness-source r2` a target is mandatory, so the orphan rule could never fire. What was wrong is that `--r2-connect-timeout-ms`'s own doc still claimed it "is rejected by name instead of being silently ignored", which `tuning: &[]` makes unreachable there. Corrected in 2f481c5, along with the dangling `[reject_conflicting_r2_targets]` link on `build_r2_client` you spotted.

Not taken: hoisting into `stateless-r2` as `resolve_target`. The rules reference each binary's flag spellings and produce operator-facing text, and `stateless-r2` is the leaf crate mega-reth's uploader consumes; `stateless-common` already holds the flag vocabulary and both binaries depend on it.

bin/stateless-validator/src/app.rs
flyq: Addressed, including the `zip` argument. The Access pair moved off clap into the shared post-parse pass: `validate_access_pair` (`crates/stateless-common/src/r2_args.rs`), called from `validate_r2_flags` alongside the target rules, naming both flags on a half-set pair and naming `--r2-custom-domain` when a whole pair has no domain. The emptiness sweep covers both halves, so `..._ACCESS_CLIENT_ID=` cannot become an empty header. No clap coupling attribute remains on either half in either binary.

Both `Option::zip` sites are now downstream of the rule rather than of a clap attribute — the validator validates in `build_r2_client` before zipping in the custom-domain arm, the trace server runs the same shared function on every startup well before its zip. Covered by `a_half_set_access_pair_is_named` plus binary-level assertions in both CLI test suites.

crates/stateless-r2/src/fetch.rs
```diff
@@ -184,25 +204,228 @@ pub struct FetchedObject {
...
+/// Each connection's share of the in-flight GET cap: `None` = unlimited, `Some(0)` clamps to 1
+/// — a zero-permit semaphore would wedge every fetch on `acquire()` forever.
+///
+/// The share rounds *up*, so with a cap that does not divide evenly the fetcher-wide total
+/// exceeds it by less than the connection count. Rounding down would round some connection to
+/// zero, and a connection that can carry nothing is worse than a cap overshot by a handful.
+fn per_connection_permits(max_concurrent_requests: Option<usize>, connections: usize) -> usize {
+    max_concurrent_requests
+        .map_or(Semaphore::MAX_PERMITS, |max| max.div_ceil(connections.max(1)))
+        .max(1)
```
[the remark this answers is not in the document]
flyq: Real arithmetic, and one of the two cases is now closed — but not by making the split exact.

The rounding up is deliberate and documented at the split (`fetch.rs`), in `concurrency_splits_across_connections_rounding_up`, and in AGENTS.md/README: rounding down leaves a connection at zero permits and wedges every GET routed to it, which is a worse failure than a total that overshoots by less than the connection count. The bound with actual consequences — the per-connection share against the edge's `SETTINGS_MAX_CONCURRENT_STREAMS` — is exact, and it is the same expression the startup advisory warns on. The fetcher-wide total corresponds to no protocol, edge, or memory limit.

The case you're right about is the other one: `connections > max`, where the multiplication stops being a rounding residue. That is now rejected at startup by name, naming both flags (2f481c5) — consistent with how this PR handles the analogous S3 case, and rejected rather than clamped because clamping would hand back fewer connections than the `r2_connections` gauge reports. It is also nonsense on its own terms: more connections than permits leaves some of them permanently idle.

The two clap help strings that said "split evenly" without the "rounded up" qualifier the prose docs carry are corrected in the same commit.

crates/stateless-r2/src/fetch.rs
```diff
@@ -290,13 +729,15 @@ impl R2ObjectFetcher {
...
+                let connection = self.next_connection();
                 let queued = Instant::now();
                 // A deadline bounds the queue wait too: under saturation the caller's budget
                 // must stay available for its fallback, not drain waiting for a permit.
                 let permit = match deadline {
-                    None => self.concurrency.acquire().await,
+                    None => connection.permits.acquire().await,
                     Some(d) => {
-                        match tokio::time::timeout_at(d.into(), self.concurrency.acquire()).await {
+                        match tokio::time::timeout_at(d.into(), connection.permits.acquire()).await
```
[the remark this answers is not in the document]
flyq: The mechanism is described correctly — the connection is picked from the round-robin cursor before its semaphore is awaited, and a queued request cannot migrate. Not changing it here, for a reason that came out of trying to write the fix.

First-fit from the cursor only differs from plain round-robin when the cursor's connection is full, i.e. under saturation. Under saturation the connection with free permits is, in the fault this flag exists for, precisely the one that just dropped every GET riding it — they all failed and released their permits together. So capacity-aware selection steers the retry into the fault it is meant to escape; least-loaded has the same property. In the unsaturated regime it degenerates to what is there now. Either inert or self-defeating on the one availability property the flag is for.

On reachability and blast radius: `--r2-connections` defaults to 1 in both binaries, where this is bit-for-bit the previous single-queue behaviour, and it needs an explicit concurrency cap to bite at all — the validator's defaults to unlimited, which is `Semaphore::MAX_PERMITS`, so nothing queues on a permit. On the trace server the R2 attempt runs on a sub-deadline (half the remaining witness stage, an eighth near tip) and any failure including `Deadline` falls through to the RPC chain; `fetch_witness_hung_r2_leaves_budget_for_the_rpc_fallback` already pins that for the strictly worse fully-hung case. And against the N=1 baseline this replaces, the same degraded-transport fault costs ~1/N of requests instead of 100%.

What would actually address it is health-aware routing — deprioritise a connection on recent failures rather than on free capacity — which needs failure data from a soak at N>1 that we do not have. Logged as a follow-up rather than done blind.

bin/stateless-validator/src/r2_witness.rs
```diff
@@ -186,6 +240,7 @@ impl R2WitnessClient {
                 trace!(number, "R2 witness fetched and decoded");
                 // Queue wait on the self-imposed concurrency cap is subtracted: folded in, it
                 // would masquerade as R2 slowness.
+                self.publish_negotiated_version();
```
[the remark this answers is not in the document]
flyq: Confirmed and fixed in 2f481c5, with one correction to the framing: the fetcher already latches the version on any HTTP response — `get_object` calls `observe_version` immediately after `send()`, before any status classification — so 403s and 404s were observed. The gap was entirely in the adapters, and it was wider than the decode case on the validator: it published only from the decode-success arm, so a GET failure never published either. The trace server published after the GET succeeded, so it covered decode failures but not 403s or frontier 404s.

Both now publish immediately after the fetcher returns, on every outcome — GET error, decode error, decode panic, success — which is the case where the answer is actually being looked for.

crates/stateless-r2/src/fetch.rs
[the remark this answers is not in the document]
flyq: Reversing this — you were right, and my reasoning above only covered half of it.

The `/simplify` pass on this branch reached the same finding independently, from an angle my reply did not address: work conservation. Committing to the cursor's connection and then waiting on its semaphore partitions one budget of `max` into `N` budgets of `max/N`, which queues distinctly worse at the same offered load regardless of whether anything is faulty. That is not a fault-tolerance argument, so the "under saturation the free connection is the one that just failed" objection I raised does not touch it. And it bites at the configuration this flag was added for: production runs a cap of 48, so `--r2-connections=8` means 6 permits per connection instead of one pool of 48.

Re-examining the fault case too, my objection was weaker than I made it sound. A connection that is *slow but alive* holds its permits, so a capacity-aware search routes away from it — exactly right. A connection that *dropped* has free permits, so the search does route there, but that costs a fast connect error and a retry that rotates off it, not a stall. The status quo was the worse of the two.

Fixed in 466b813: `ConnectionPool::try_acquire` searches from the rotating cursor for a connection with a free permit. Only when every connection is full does a fetch wait, and it waits on the cursor's own pick rather than on whichever has the most room — that part of the original objection does hold, so capacity is not used to choose *what to wait on*. Pinned by `a_saturated_connection_is_skipped_while_another_has_room` in `fetch.rs`.

crates/stateless-r2/src/fetch.rs
```diff
@@ -184,25 +204,288 @@ pub struct FetchedObject {
...
+///
+/// `label` and `example` name the flag the caller owns, so each target keeps its own error.
+/// The raw input is never echoed back: a rejected shape may carry inline credentials
+/// (userinfo), and this message ends up in startup logs.
+fn parse_target_origin(
+    input: &str,
+    label: &str,
+    example: &str,
+) -> Result<(String, String), String> {
+    let (origin, host) = parse_endpoint(input);
```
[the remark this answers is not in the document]
flyq: Not reproducible — `Url::host_str` keeps the brackets on an IPv6 host, so the rebuilt origin is `http://[::1]:8080`, not `http://::1:8080`.

Checked against the parser rather than by reading: `Url::parse("http://[::1]:8080").host_str()` returns `"[::1]"` (and `port()` returns `8080` separately), so `parse_endpoint`'s `format!("{host_str}:{port}")` produces `[::1]:8080` and the origin reassembles intact. `is_loopback_host` then strips the brackets on its own `[` branch and parses `::1`, which is `Ipv6Addr::LOCALHOST`.

Pinned in eed59a8 rather than just answered here: `ipv6_loopback_origins_survive_the_round_trip` builds a fetcher for `http://[::1]:8080` and `http://[::1]` with Access credentials — both must be accepted, and `origin()` must come back bracketed — and asserts that `http://[2001:db8::1]:8080` is still refused for carrying credentials over plaintext to a non-loopback host. Nothing covered that round trip before, and it spans two modules, so a future change to how the origin is reassembled would have broken the port-forward path with an error that names exposure.

Troublor (approved):

vincent-k2026 (approved):

PR #194, opened 2026-09-03
refactor: behavior-preserving cleanups and unused-API removals

flyq:
## Summary

PR 1/6 of the #170 split. Everything here is behavior-preserving on canonical-block paths, plus the public-API removals that had no consumer left. If a bisect ever lands on this PR, that is a strong signal something in it was **not** behavior-preserving — which is the whole point of separating it from B–F.

Responds to [vincent's review of #170](https://github.com/megaeth-labs/stateless-validator/pull/170#pullrequestreview-4852987751): *"Suggest splitting into a genuinely behavior-preserving cleanup plus a handful of small, individually-motivated changes."* Every item he listed under **"Worth keeping, whatever the split"** is in here.

## What's in it

**Second sources of truth removed**
- `ChainStore` drops the `ContractStore` supertrait — nothing consumed contracts through it. The forced stubs on the pipeline mock and the trace server's `StubBlockStore` go with it.
- `BlockMeta::from_header` / `try_from_header` replace six hand-rolled header→meta projections: the three tip observations (validator + two in the trace server) default a missing `withdrawals_root` to zero, and the three anchor-init sites (`app.rs`, the trace server's `main.rs`, the validator integration test) go through the strict variant, which returns `None` so each keeps its own reject error — that path must reject, not default.
- `FetcherState::in_flight_blocks` deleted: it mirrored `task_to_block.values()`. Rather than scan those values per completion (O(`max_in_flight`²), Codex's P2), the one remaining map is keyed by block — `in_flight: HashMap<u64, Id>` — so `recover_gaps` and the success/failure paths stay O(1); the task id is only needed to map a panicked task's `JoinError` back to its block, a scan on that rare path.
- The hardfork reorder loop in `ChainSpec::from_genesis` is gone, and so is the `mega_mainnet_hardforks()` ladder: both were copies of mega-evm's `MegaHardfork` declaration, which the chain-spec tests now read directly as `MegaHardfork::VARIANTS`. The "unknown fork" append branch was unreachable — `into_vec` only ever produces those ten.
- `BackoffSchedule` (stepped from `RetryPacing::schedule()`) collects the jittered-doubling arithmetic (jitter, `max` cap, 1 ms floor) that was duplicated between the RPC round-robin loop and the R2 GET loop, comments and all. Both consume it — and `BackoffPolicy` itself is now a re-export of `RetryPacing`: the two were field-for-field identical `{initial, max}` pairs with a hand-written adapter between them (`r2_witness.rs::pacing`), which goes too.
- The validator's integration mock handlers dedupe onto `MockServerState::block_by_number_hex` / `block_by_hash` / `invalid_params` — which also fixes a drifted bare `unwrap()` on malformed `eth_getHeaderByNumber` params.

**Dead parameters, accessors, and backdoors**
- `handle_transient_restart`'s `_reason: String` parameter and the upstream string construction that fed it.
- `ValidatorFetcher::on_remote_height` fn-pointer injection (the only value ever passed was `metrics::set_remote_chain_height`).
- `run_with_signals` no longer takes the report endpoint (an `Option<String>` it only ever `.is_some()`'d); it reads `RpcClient::reports_validation()`, since the client is already built from the same flag — two inputs that had to agree became one.
- `LightWitnessExecutor::kvs()` — no consumer, here or in mega-reth.
- `#[cfg(test)] ValidatorDB::set_anchor_block` — the roundtrip test now goes through the production `reset_to_anchor` path instead of a test-only shortcut around the helper layer.
- `WitnessExternalEnv::new` / `from_light_witness` share one `from_metadata_kvs` body (the metadata scan was duplicated verbatim); a doubled doc block on `create_evm_env` removed.

**Reuse**
- The blob base-fee fraction comes from revm's `BLOB_BASE_FEE_UPDATE_FRACTION_CANCUN` instead of a hand-copied consensus constant. Verified equal: revm-primitives defines it as `3_338_477`, the local constant was `3338477`.
- `install_prometheus_exporter` hoisted into `stateless_common::metrics` behind a `prometheus-exporter` feature both binaries enable; the `metrics-exporter-prometheus` dep moves from the two bins into common as an optional dep, so mega-reth (which consumes `stateless-common` and pins the exporter at 0.16) does not build a second copy for a function it never calls.
- The validator's contract-hash collection uses the existing `collect_code_hashes` instead of open-coding `iter_code_hashes().collect::<HashSet<_>>()`.
- `workers.rs` → `runner.rs` (it runs the pipeline; the workers live in core), docs updated.

## The invariant that lost its only enforcement

vincent: *"Removing `mega_mainnet_hardforks()` and the explicit reordering loop makes `MegaethGenesisHardforks::into_vec`'s literal the single source of activation order. That is the right direction… But **nothing tests the order.**"*

`test_mega_hardforks_iterate_in_activation_order` now pins `into_vec` end-to-end through `from_genesis` against mega-evm's own declaration rather than a local ladder: it schedules every `MegaHardfork::VARIANTS` entry (genesis field derived from the variant name) and asserts the scheduled MegaETH forks equal `VARIANTS`, so a variant `into_vec` lacks, duplicates, or misplaces fails at either end of the ladder — a hand-written expected list or a prefix compare both let a trailing omission through, and new forks are only ever appended. Mutation-checked on the final tree: dropping Rex6 from `into_vec` fails, swapping Rex5/Rex6 fails, duplicating Rex6 fails. `mainnet_genesis_schedules_every_canonical_hardfork` iterates `VARIANTS` the same way, through the same name-keyed `ChainHardforks::fork` lookup fork selection uses at runtime, which is what let the hand-written ladder go. One correction to the earlier framing: insertion order is *not* consensus-relevant. mega-evm resolves the active fork through `mega_fork_activation`, a name-keyed map lookup on `ChainHardforks`, and nothing outside the tests iterates `forks_iter()`, so the deleted reorder loop was already inert for consensus; what the tests guard is that the local literal and mega-evm's declaration agree in membership and order, so a fork the executor supports cannot go unscheduled unnoticed.

## Deviations from the #170 material

One, forced by main moving 9 PRs since `e1b2ede`:

1. **`BackoffSchedule` lives in `stateless-r2`, not `stateless-common`.** #187/#192 moved the R2 GET loop into `stateless-r2`, and `stateless-common` depends on `stateless-r2` — so common is no longer a home both consumers can reach, and `stateless-r2` must stay free of upward dependencies (its `RetryPacing` doc says so explicitly). The schedule therefore sits next to `RetryPacing` as `RetryPacing::schedule()`, and `BackoffPolicy` is `RetryPacing` re-exported under the name `stateless-common`'s API uses (`RetryPacing` gained the `const fn new` that `BackoffPolicy` had; every existing constructor, struct-literal, and field-access form compiles unchanged). For mega-reth this is an *addition* to `stateless-r2`'s surface, not a break.

## Observable changes (behavior-preserving ≠ byte-identical output)

- The validator's `/metrics` loses the `eth_getTransactionByHash` counter series. It was pre-registered but permanently zero — that method is only ever called by the trace server.
- The trace server's exporter-install failure message changes from "Failed to install metrics exporter" to "Failed to install Prometheus exporter" (the validator's wording), since both now share one installer. Startup-only.
- `collect_code_hashes` returns a sorted, deduplicated `Vec` where the open-coded `HashSet` path was unordered. No consumer depended on the order; the cache lookup and the RPC batch are both order-insensitive.

## Breaking at the next tag

mega-reth pins `stateless-core` at `v2.0.14`, so nothing breaks until the bump — which needs coordination:

- `LightWitnessExecutor::kvs()` removed.
- `BLOB_GASPRICE_UPDATE_FRACTION` removed; use revm's `BLOB_BASE_FEE_UPDATE_FRACTION_CANCUN` (same value, verified — it is a consensus parameter).
- `ChainStore: ContractStore` supertrait dropped. Downstream payoff: mega-reth's `MegaValidatorChainStore` can delete the `ContractStore` impl its own comment calls "intentionally inert".
- `stateless_common::BackoffPolicy` is now a re-export of `stateless_r2::fetch::RetryPacing` — source-compatible (same field names, `new`, and it gains `Copy`); only a trait impl naming the type would notice, and mega-reth has none. `Copy` also turns any `.clone()` on it into a `clippy::clone_on_copy` warning (this PR dropped two of its own on the retry field); the mega-reth branch carrying the validator crates has no `BackoffPolicy` reference today, but grep before the pin bump.
- `mega_mainnet_hardforks()` removed — the #170 deletion, restored now that its one consumer reads `MegaHardfork::VARIANTS`. No consumer in mega-reth.

## Testing

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features` (0 warnings), `cargo sort --check`, full workspace suite **478 passed / 0 failed**, and `cargo test -p stateless-core --no-default-features --lib --no-run` (the no-std alias path) all clean.

## Notes

PR 1 of 6, stacked: B (EIP-3155 writer deletion) → C (verify/gas_used) → D (blocking work) → E (hot-path allocations) → F (empty witness endpoints). Each targets its predecessor and stands on its own.


crates/stateless-core/src/pipeline/fetcher.rs
```diff
@@ -129,8 +124,8 @@ impl<F: BlockFetcher> FetcherState<F> {
         let mut recovered = 0;
         for bn in self.base_block..self.next_block {
             if !self.sent.contains(&bn) &&
-                !self.in_flight_blocks.contains(&bn) &&
-                !self.failed.contains(&bn)
+                !self.failed.contains(&bn) &&
+                !self.task_to_block.values().any(|&in_flight| in_flight == bn)
```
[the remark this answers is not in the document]
flyq: Fixed in 5918ac5. `FetcherState` now keys its one in-flight map by block instead of by task id (`in_flight: HashMap<u64, Id>`, `crates/stateless-core/src/pipeline/fetcher.rs:28`), so `recover_gaps` (`:131`), `on_success` (`:93`) and `on_failure` (`:98`) are O(1) per block again without reinstating the mirror set this PR removed. The task id is only needed to map a panicked task's `JoinError` back to its block, and that lookup scans `in_flight` on the panic path alone (`:106`).

vincent-k2026 (commented):
Right split of #170: second sources of truth and dead parameters go, hardfork order moves from a runtime reorder to a test-time cross-check. No correctness/security blocker.

## What's good

- `ChainStore` no longer extends `ContractStore` — the pipeline mock and `StubBlockStore` stubs go with it, and mega-reth can drop the "intentionally inert" impl on the next pin bump.
- `BlockMeta::from_header` (tip: missing `withdrawals_root` → zero) vs `try_from_header` (anchor: `None`, each site keeps its own reject error) — defaulting and rejecting are not collapsed into one constructor.
- Fetcher `in_flight: HashMap<u64, Id>` fixes the Codex P2 O(n²) `recover_gaps` at the source, without bringing the mirror set back.
- `BackoffPolicy` as a `RetryPacing` re-export plus `RetryPacing::schedule()`: jitter / max / 1 ms floor live in one place; RPC and R2 GET no longer duplicate the arithmetic; `stateless-r2` stays free of upward deps.
- `mega_mainnet_hardforks()` kept as the membership oracle (#190 already consumes it); only the reorder loop goes. Matches main having moved.
- `test_mega_hardforks_iterate_in_activation_order` is the deleted runtime cross-check, moved to test time. The claimed Rex5/Rex6 swap going red is the right shape.
- `set_anchor_block` now goes through production `reset_to_anchor`; the mock `eth_getHeaderByNumber` `parse().unwrap()` is `parse()?`.
- Breaking surface is explicit (`kvs()`, `BLOB_GASPRICE_UPDATE_FRACTION`, `ChainStore` supertrait). mega-reth is pinned at `v2.0.14`, so nothing breaks until the coordinated bump.

## Blocking (correctness / security)

None.

## Non-blocking / follow-up

1. The order test filters `forks_iter()` by ladder names, so an extra fork in `into_vec` is silently dropped. Omitting it from the ladder fails; adding it only to `into_vec` does not. `assert_eq!(mega_order.len(), expected.len())` (or comparing the MegaETH prefix without a filter) would make that symmetric. Today it is a comment: insert in both places.
2. `/metrics` loses the permanently-zero `eth_getTransactionByHash` series, and the exporter-install error string becomes the shared "Prometheus exporter" wording. Disclosed in the PR body; dashboards that keyed on the old name need to know.
3. `workers.rs` → `runner.rs` is behavior-preserving and the docs were updated; it is a rename, not an unused-API removal, so greps for the old path will wobble once.
4. `install_prometheus_exporter`'s `expect("valid bucket config")` is hoisted as-is from the two binaries, not newly introduced. Startup-only, buckets are static — fine.

flyq:
Thanks — item 1 holds and is fixed; 2–4 need no code change.

**1. Order test was one-directional.** Confirmed: `filter(|name| expected.contains(name))` drops a fork that exists only in `into_vec`, so that omission passed. Fixed by comparing the prefix instead of filtering — `from_genesis` concatenates the MegaETH forks ahead of the Optimism ones (`chain_spec.rs:167-168`), so they are the leading `ladder.len()` entries, and `take(expected.len())` makes the check symmetric (`chain_spec.rs:406-419`).

Mutation-checked both directions on the same tree:

| mutation | filter (before) | prefix (after) |
|---|---|---|
| swap Rex5/Rex6 in `into_vec` | fails | fails |
| drop Rex5 from the ladder only | **passes** | **fails** |

The second row is exactly the blind spot you named. The doc on `into_vec` already says to insert in both places; now the test enforces it.

**2. `eth_getTransactionByHash` series / exporter wording.** Agreed, both are in the PR body's "Observable changes" section so the dashboard owners see them at merge time. No code change.

**3. `workers.rs` → `runner.rs`.** Agreed it is a rename rather than a removal; `AGENTS.md` and `README.md` both point at the new path, so the wobble is one grep.

**4. `expect("valid bucket config")`.** Agreed — hoisted verbatim, startup-only, static buckets.

crates/stateless-core/src/chain_spec.rs
```diff
@@ -382,6 +381,48 @@ mod tests {
...
+        // `from_genesis` concatenates the MegaETH forks ahead of the Optimism/Ethereum ones, so
+        // they are the leading `expected.len()` entries. Taking that prefix rather than
+        // filtering by ladder membership keeps the check symmetric: a fork added to `into_vec`
+        // alone shifts the prefix and fails here, where a membership filter would have dropped
+        // it and passed.
+        let mega_order: Vec<&str> = spec
+            .hardforks
+            .forks_iter()
+            .map(|(hardfork, _)| hardfork.name())
+            .take(expected.len())
```
RealiCZ: `take(expected.len())` closes the middle-of-ladder case you mutation-checked (drop Rex5 from the ladder only), but not the trailing one: a fork appended to `into_vec` that the ladder lacks lands *after* the prefix, so `mega_order == expected` still holds and the test passes. New forks are only ever appended (Rex7 in mega-evm #360 is the live example), so that is the realistic omission, and the new comment's "shifts the prefix and fails here" does not hold for it. The `len()` assert suggested earlier has the same hole.

The fix needs a membership oracle that is not the ladder. `MegaHardfork::VARIANTS` is already there on v1.7.0 (alloy-hardforks' `hardfork!` macro emits it; mega-evm declares `MegaHardfork` through that macro):

```rust
let is_mega = |n: &str| MegaHardfork::VARIANTS.iter().any(|v| v.name() == n);
let expected: Vec<&str> = MegaHardfork::VARIANTS.iter().map(|v| v.name()).collect();
let mega_order: Vec<&str> = spec.hardforks.forks_iter()
    .map(|(h, _)| h.name())
    .filter(|n| is_mega(n))
    .collect();
assert_eq!(mega_order, expected);
```

That fails on a missing, extra, or misordered `into_vec` entry alike, and pins the literal against the order mega-evm resolves with after #362 (`VARIANTS.iter().rev().find(..)`). If the genesis fields in this test are also generated from `VARIANTS` (`lcfirst(name) + "Time"`, plus the Rex5/Rex6 configs), a missing `MegaethGenesisHardforks` field fails here too.

crates/stateless-core/src/chain_spec.rs
```diff
@@ -227,6 +211,13 @@ impl MegaethGenesisHardforks {
     }
 
     /// Convert the MegaETH genesis hardforks into a vector of hardforks and their conditions.
+    ///
+    /// The literal below is the single source of the canonical MegaETH activation order:
+    /// [`ChainSpec::from_genesis`] merges it as-is, and fork selection by timestamp walks
+    /// `forks_iter()` in insertion order, so a wrong order here means wrong hardfork params
```
RealiCZ: This rationale is not what the code does. `create_block_execution_env` resolves the fork through mega-evm's `MegaHardforks::hardfork(timestamp)`, which on v1.7.0 is an explicit Rex6 → MiniRex if-else chain and after mega-evm #362 is `MegaHardfork::VARIANTS.iter().rev().find(..)`. Both go through `mega_fork_activation(fork)`, which here is `self.hardforks.fork(fork)`, a name-keyed map lookup in reth's `ChainHardforks`. Insertion order in `ChainHardforks` never enters spec or `BlockLimits` selection; the only `forks_iter()` consumers in this repo are tests. So the removed reorder loop was already inert for consensus (which makes the removal safer than the PR argues), and a wrong `into_vec` order cannot produce wrong limits.

Please reword this and the `mega_mainnet_hardforks` doc (`:297`) to what the tests actually guard: that the local literal and mega-evm's declaration agree in membership and order, so a fork the executor supports cannot go unscheduled unnoticed. With the VARIANTS-based test above, the hand-written ladder is a second copy of that declaration; #362 removed the equivalent copy on the mega-evm side for the same reason, and `mainnet_genesis_schedules_every_canonical_hardfork` can iterate `VARIANTS` directly. Dropping the test-name references from the docs would also keep them from rotting on a rename.

RealiCZ (commented):
Two follow-ups on the hardfork section, both inline. Also for the "Breaking at the next tag" list: `BackoffPolicy` gaining `Copy` means any `.clone()` on it in mega-reth becomes a `clippy::clone_on_copy` warning, which this PR itself had to remove here; worth a grep before the pin bump.

crates/stateless-core/src/chain_spec.rs
flyq: Holds — the prefix compare had exactly that trailing hole. Fixed in f43fdfe: `test_mega_hardforks_iterate_in_activation_order` schedules every `MegaHardfork::VARIANTS` entry (genesis field derived from the variant name, `crates/stateless-core/src/chain_spec.rs:371-383` and `:409-412`) and asserts the scheduled MegaETH forks equal `VARIANTS` (`:388-403`), so a missing, duplicated, or misordered `into_vec` entry fails at either end of the ladder. Mutation-checked on the new tree: drop Rex6 from `into_vec` → fails; swap Rex5/Rex6 → fails; duplicate Rex6 → fails.

crates/stateless-core/src/chain_spec.rs
flyq: Agreed, and verified against the pinned sources: mega-evm v1.7.0's `MegaHardforks::hardfork` is the Rex6 → MiniRex if-else over `is_*_active_at_timestamp`, each of them `mega_fork_activation(fork).active_at_timestamp(ts)`, and reth v1.6.0's `ChainHardforks::fork` is `self.map.get(fork.name())`. Insertion order is never consulted, and `forks_iter()` has no non-test consumer in this repo.

Fixed in f43fdfe:
- `into_vec` doc reworded to that: fork selection never reads the order; what the tests guard is agreement with mega-evm's declaration in membership and order (`crates/stateless-core/src/chain_spec.rs:213-221`).
- `mega_mainnet_hardforks()` removed. `mainnet_genesis_schedules_every_canonical_hardfork` iterates `MegaHardfork::VARIANTS` through the same `ChainHardforks::fork` lookup (`:723-740`), with `NOT_YET_SCHEDULED` typed as `&[MegaHardfork]` (`:718`).
- Test-name references dropped from the docs.

PR body updated to match: the "order is consensus-relevant" claim is retracted, and `mega_mainnet_hardforks()` joins the breaking list (no mega-reth consumer).

flyq:
On the `BackoffPolicy: Copy` note — added to the "Breaking at the next tag" list. This PR dropped two `.clone()`s of its own on the retry field; the mega-reth branch carrying the validator crates has no `BackoffPolicy` / `rpc_retry` reference today, so nothing to change there yet, but the grep is on the pin-bump checklist.

RealiCZ (approved):
f43fdfe closes both hardfork threads: the order test now pins `into_vec` against `MegaHardfork::VARIANTS` (symmetric on a missing, extra, duplicated, or misplaced entry), the membership test iterates `VARIANTS` through the same name-keyed lookup fork selection uses, the hand-written ladder is gone, and the `into_vec` doc states what the tests actually guard. The two tests fail in complementary ways on a mega-evm bump that brings a new fork, so the next pin bump cannot leave it unscheduled or unmapped unnoticed.

Troublor (approved):

PR #196, opened 2026-09-03
fix: check each RPC transaction hash against its own encoding in verify_block_integrity

flyq:
## Summary
PR 3/6 of the #170 split, stacked on #195. One rewrite of a security-critical check, separated from the cleanup so it gets argued on its own. It turned out to be a fix rather than a perf change: the check it replaces was a tautology.

## `verify_block_integrity` checks the claimed transaction hash against the envelope's own encoding
Before: `tx.inner.clone().into_inner().trie_hash() == *hash()` per transaction, then a second clone inside `ordered_trie_root_with_encoder` for the transactions root.

That hash check never checked anything. `Encodable2718::trie_hash()` defaults to `keccak256(encoded_2718())`, but every variant in play overrides it: `Signed<T>::trie_hash` returns `*self.hash()` (alloy-consensus 1.1.2 `src/signed.rs:515`), `Sealed<T>::trie_hash` returns `self.hash()` (alloy-eips 1.1.2 `src/eip2718.rs:327`), and `OpTxEnvelope`'s `TransactionEnvelope` derive forwards per variant. The deserializer seeds that cached hash from the RPC `hash` field through `new_unchecked` (`src/signed.rs:583`), so the old predicate compared the provider's claim with itself. A provider returning the right transaction bytes under a wrong `hash` passed verification, and the transactions-root check cannot catch that because the root is computed from the bytes.

After: each envelope is encoded once via `encoded_2718()`; `keccak256` of those bytes is checked against the claimed hash, and the same bytes feed the ordered trie (`crates/stateless-common/src/rpc_client.rs:1587`, `:1609`). The two clones per transaction are gone and the keccak is new work, the first real check of the field. The other half of the trade is memory: `encoded_txs` keeps every transaction's encoding resident until the trie root finishes, where the old encoder closure reused one scratch buffer per item — one wire-size copy per transaction, on top of the parsed `Block<Transaction>` that is resident anyway.

**Pinned**: `verify_block_integrity_rejects_a_forged_transaction_hash` (`rpc_client.rs:3001`) forges the `hash` of the first and last transaction of a mainnet fixture block through a JSON round trip and asserts the block is rejected, with the transactions root left intact so only the hash check can reject. It asserts index 0 is the deposit (`Sealed` override) and the last index a signed envelope (`Signed` override), so a fixture change cannot silently drop either path.

## Testing
`cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features` (0 warnings), `cargo sort --check`, full workspace suite **481 passed / 0 failed**, `cargo test -p stateless-core --no-default-features --lib --no-run` clean.

The positive path is covered by the validator's mock-RPC integration test, which fetches every fixture block through the verifying path.


vincent-k2026:
Checked both claims against the upstream sources (alloy-consensus / alloy-eips 1.1.2, alloy-evm's trait defaults, mega-evm `v1.7.0` `30ce038`). Part 1 is a real fix and the test is the right shape. Part 2 is equivalent as argued, but I'd drop it. No blockers.

## Part 1: the tautology is real

Verified each link:

- default `Encodable2718::trie_hash()` is `keccak256(encoded_2718())` — alloy-eips `eip2718.rs:272-274`
- `Signed<T>` overrides it to `*self.hash()` — alloy-consensus `signed.rs:515-517`
- `Sealed<T>` (the deposit path) overrides it to `self.hash()` — alloy-eips `eip2718.rs:327-329`
- the cached hash is seeded straight from the RPC `hash` field — `signed.rs:578-588`, `Signed::deserialize` -> `new_unchecked(tx, sig, value.hash.into_owned())`

So the old check compared the provider's claim with itself, and the transactions-root check can't cover for it since the root comes from the bytes. `verify_block_integrity` is the trust boundary against the RPC provider (`rpc_client.rs:544`), so a check that reads like it validates the hash and never did is worth calling a fix — agreed on the retitle.

`verify_block_integrity_rejects_a_forged_transaction_hash` is the right shape: positive control first, then both override paths (`Sealed` at index 0, `Signed` at the end), with the transactions root left intact so only the hash check can reject. No need for the "I reverted it locally" evidence — the four source facts above already imply it.

## Part 2: equivalent, but the change buys nothing

One link the description skips: the call site is `executor.apply_post_execution_changes()` (`executor.rs:488`), not `finish()`. They're connected by alloy-evm's default impl (`block/mod.rs:209-216`, `self.finish().map(|(_, result)| result)`), which `MegaBlockExecutor` doesn't override, and the factory here really is `MegaBlockExecutorFactory` (`executor.rs:293`). Worth a sentence, because if mega-evm ever overrides `apply_post_execution_changes` the argument breaks silently.

With that, the equivalence holds — `finish()` is character-for-character the old expression, computed after `commit_system_call_outcomes`.

My problem is the ROI:

- **No benefit.** `receipts.last()` is O(1). No perf win, no fewer lines. The only difference is where the value comes from.
- **The dependency points the wrong way.** `header.gas_used` *is* "the last receipt's cumulative gas" by consensus definition. The old line was that definition; the new one trusts an executor's self-reported field, backed by a comment and a version pin. Concretely: the receipt chain is anchored to the header through the receipts-root check in `verify_replay_outputs`; `BlockExecutionResult::gas_used` is anchored by nothing.
- **A drift shows up as a false reject, not a miss.** If upstream ever redefines the field, every legitimate block fails `GasUsedMismatch` and it reads like a chain-wide consensus divergence. Whether CI catches it first depends on a fixture happening to cover the differing case.

On dropping the `debug_assert_eq!` — the reasoning ("it could only fire on a *correct* upstream redefinition, which the header check adjudicates") covers half of it. If the redefinition is *wrong*, the header check turns everything red and the assert is what tells you it's a field-semantics drift rather than a real divergence. One adjudicates, the other attributes; they're not the same altitude.

Suggestion: either drop this hunk (the old line is the spec written out) or keep it and restore the assert. I'd lean to dropping it — part 1 is a clean fix and this is the only hunk in the PR that needs an argument at all.

## Small stuff

1. `tx.inner.inner.inner()` — three `inner`s and no way to tell what each layer is (op `Transaction` -> alloy `Transaction` -> `Recovered::inner()`). A named binding would help.
2. The description says the two clones are gone but doesn't mention that `encoded_txs: Vec<Vec<u8>>` now holds the whole block's encodings resident — not free on a high-TPS block. Fair trade, just say both halves.
3. The test's coverage claim ("first is the deposit (`Sealed`), last a signed envelope") is a comment, not an assertion. If a fixture change makes index 0 a normal tx, the test stays green and the `Sealed` path silently loses coverage. Cheap to assert the variant.
4. Commit headline is still `perf: encode each transaction once...` (`e62e3e14`). If this isn't squashed under the PR title, a security fix lands in the log as a perf change and gets hard to find later.
5. The bot's open question is stale — the current description already explains why the `debug_assert_eq!` and the third fixture test were dropped. A one-line reply closes it.

Troublor (approved):

flyq:
Dropped the `gas_used` hunk: `crates/stateless-core/src/executor.rs` is back to the base version in 5c39a74, so the header check reads the last receipt's cumulative gas as before and the PR is the hash check alone. Title and description updated to match.

Small items, same commit:
1. The envelope comes out of a named `Recovered<_>` binding (`crates/stateless-common/src/rpc_client.rs:1585`), and the signer binding is `recovered_signer` so the two cannot be read as one another.
2. Description states both halves of the trade: the two clones are gone, and `encoded_txs` holds one encoded copy per transaction until the trie root finishes.
3. The test asserts the variant at each index (`rpc_client.rs:3018`): index 0 `is_deposit()`, the last index not.
4. The repo allows squash merges only and takes the squash title from the PR title (ruleset `allowed_merge_methods: ["squash"]`, repo `squash_merge_commit_title: PR_TITLE`), so `fix:` is what lands in the log; the branch commit's `perf:` headline never reaches `main`.
5. Answered in a separate comment so the bot's reconcile round picks it up.

flyq:
Answering the open question: the `gas_used` hunk is gone in 5c39a74 (`crates/stateless-core/src/executor.rs` is back to the base version, reading the last receipt's cumulative gas), so there is no derivation-site pin to restore and the description no longer mentions one. `verify_replay_outputs` (`executor.rs:622`) is unchanged.

RealiCZ (approved):

PR #207, opened 2026-09-08
ci: attach the release binaries to the GitHub Release

Troublor:
The extension half for stateless-validator — assets only, as requested.

- `on-release.yml` on `release: published`: builds `stateless-validator` and `debug-trace-server` at the released tag (same commands as `release.yaml` / `release-tracing.yaml`, one `cargo build --release --locked`), reports their sizes and `--version` output, and attaches both plus a generated `SHA256SUMS` to the Release page (`release-assets`, `--clobber` so re-runs replace).
- `workflow_dispatch` on a tag ref with `dry_run=true` (default) rehearses the build and checksums and attaches nothing: `gh workflow run on-release.yml --ref vX.Y.Z -f dry_run=true`. The tag is always the run's own ref — no separate input.
- No credentials involved (job token, `contents: write`), so no environment is needed. The Artifact Registry archive stays exactly as it is in `release.yaml` and `release-tracing.yaml`, which the same tag push fires.

The version-report step is informational: `stateless-validator`'s CLI has no clap version, so it cannot be the gate it is for `mega-evme`.

### DIFF
--- a/.github/workflows/on-release.yml
+++ b/.github/workflows/on-release.yml
@@ -7,6 +7,14 @@ name: On Release
 #   gh workflow run on-release.yml --ref vX.Y.Z -f dry_run=true
 # The tag being released is always the run's own ref; there is no tag input.
 #
+# This file must exist on the release branch: a `release` event resolves the
+# workflow from the tagged commit's tree, not from the default branch, and the
+# tag sits on the release branch. A branch cut from a default branch that has
+# this file carries it; a branch cut before it landed needs it cherry-picked
+# (via its own PR — the `release-*` ruleset requires one) before the settle PR
+# merges, or the Release publishes with nothing attached and no failed run.
+# A rehearsal likewise only works on a tag whose tree has this file.
+#
 # The internal archive to Artifact Registry stays in release.yaml and
 # release-tracing.yaml, which the same tag push fires.
 
@@ -20,6 +28,13 @@ on:
         required: false
         type: boolean
         default: true
+
+# One run per tag: a release run and a dispatch on the same tag both end in a
+# `--clobber` upload (delete-then-upload), so overlapping runs would race.
+# Never cancel an upload mid-flight; queue instead.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
 
 permissions:
   contents: read
@@ -54,13 +69,30 @@ jobs:
       - name: Build
         run: cargo build --release --locked --bin stateless-validator --bin debug-trace-server
 
-      - name: Report binary versions
-        # Informational: stateless-validator's CLI does not carry a clap
-        # version, so this cannot be a gate the way it is for mega-evme.
+      - name: Verify the binaries
+        # stateless-validator carries a clap version (`#[clap(version)]`), so
+        # it is a gate: the built binary must report the tag's version, which
+        # also catches a tag / Cargo.toml mismatch nothing else checks here.
+        # debug-trace-server has no version flag; `--help` proves it loads and
+        # runs (a missing library or a crash fails the step) without gating.
         run: |
+          set -euo pipefail
+          expected="stateless-validator ${TAG#v}"
+          actual="$(target/release/stateless-validator --version)"
+          if [[ "$actual" != "$expected" ]]; then
+            if [[ "$DRY_RUN" == "true" ]]; then
+              # A rehearsal reports what a real run would refuse, and carries on.
+              echo "::warning::built binary reports '$actual', expected '$expected' — a real release would stop here"
+            else
+              echo "::error::built binary reports '$actual', expected '$expected'; refusing to publish a mislabeled binary"
+              exit 1
+            fi
+          fi
+          target/release/debug-trace-server --help > /dev/null
           for b in stateless-validator debug-trace-server; do
-            echo "$b: $(target/release/$b --version 2>/dev/null || echo '<no --version>')  ($(stat -c %s target/release/$b) bytes)"
+            echo "$b: $(stat -c %s target/release/$b) bytes"
           done
+          echo "$actual"
 
       - uses: megaeth-labs/.github/.github/actions/release-assets@main
         with:
--- a/.github/workflows/on-release.yml
+++ b/.github/workflows/on-release.yml
@@ -1,0 +1,71 @@
+name: On Release
+
+# Publish target for a stateless-validator release: the two binaries as
+# downloads on the GitHub Release page, with a SHA256SUMS file. Runs when the
+# release pipeline publishes the Release (release-publish creates it after
+# the tag), and on demand for a rehearsal on a tag ref:
+#   gh workflow run on-release.yml --ref vX.Y.Z -f dry_run=true
+# The tag being released is always the run's own ref; there is no tag input.
+#
+# The internal archive to Artifact Registry stays in release.yaml and
+# release-tracing.yaml, which the same tag push fires.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Build and checksum everything; attach nothing"
+        required: false
+        type: boolean
+        default: true
+
+permissions:
+  contents: read
+
+env:
+  TAG: ${{ github.ref_name }}
+  # A release event is never a dry run; a dispatch defaults to one.
+  DRY_RUN: ${{ github.event_name == 'workflow_dispatch' && inputs.dry_run || false }}
+
+jobs:
+  binaries:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 60
+    permissions:
+      contents: write # release-assets attaches to the Release
+    steps:
+      - name: Require a tag ref
+        run: |
+          [[ "$GITHUB_REF_TYPE" == "tag" ]] || { echo "::error::this workflow publishes the run's own ref, which must be a tag (got $GITHUB_REF_TYPE $GITHUB_REF_NAME); dispatch it with --ref vX.Y.Z"; exit 1; }
+
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          ref: ${{ env.TAG }}
+          persist-credentials: false
+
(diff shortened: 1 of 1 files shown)


.github/workflows/on-release.yml
```diff
@@ -0,0 +1,71 @@
...
+# the tag), and on demand for a rehearsal on a tag ref:
+#   gh workflow run on-release.yml --ref vX.Y.Z -f dry_run=true
+# The tag being released is always the run's own ref; there is no tag input.
+#
+# The internal archive to Artifact Registry stays in release.yaml and
+# release-tracing.yaml, which the same tag push fires.
+
+on:
+  release:
+    types: [published]
```
flyq, line 23: **correctness** · `CONFIRMED`

For `release` events GitHub reads the workflow file from the tag's tree, and `release-v2.0.18` — already cut at `f45133e`, before this PR — does not contain this file, so the release it was written for will publish with no binaries and no failed run.

<details>
<summary>Failure scenario</summary>

The docs list the events that require the file on the default branch (`schedule`, `workflow_dispatch`, `workflow_run`, `repository_dispatch`, …); `release` is absent from it, and its `GITHUB_SHA` is documented as "Last commit in the tagged release" — so the file is resolved at the tag. `release-publish` tags the settle merge commit on `release-v2.0.18`, and `git ls-tree -r origin/release-v2.0.18 -- .github/workflows/` lists `release-candidate` / `release-publish` / `release-settle` / `release.yaml` / `release-tracing.yaml` / `build-and-test` / `claude` / `coverage` — no `on-release.yml`. The Release publishes with nothing attached: no run, no failure, no signal that anything was skipped.

`release-publish.yml:9-10` already documents exactly this constraint for itself ("This file must exist on the release branch (it does, when the branch is cut from a default branch that has it)"). `on-release.yml` inherits the same constraint but neither states nor satisfies it — and `release-v2.0.18` is the case the parenthetical does not cover, because the branch was cut before this PR landed.

The same root cause breaks the rehearsal advertised at line 7: `gh workflow run on-release.yml --ref vX.Y.Z` resolves the workflow at that ref, and no tag that exists today (`v2.0.17` and older) contains this file, so it 404s for every currently-taggable version.

Fix: land this file on `release-v2.0.18` before the settle PR merges — the `release-*` ruleset requires a reviewed PR, so it needs its own PR against that branch, otherwise the settle merge commit (which is what gets tagged) still won't carry it. Worth adding the release-branch requirement to the header comment too, the way `release-publish.yml` does.

</details>

.github/workflows/on-release.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Build
+        run: cargo build --release --locked --bin stateless-validator --bin debug-trace-server
+
+      - name: Report binary versions
+        # Informational: stateless-validator's CLI does not carry a clap
+        # version, so this cannot be a gate the way it is for mega-evme.
+        run: |
+          for b in stateless-validator debug-trace-server; do
+            echo "$b: $(target/release/$b --version 2>/dev/null || echo '<no --version>')  ($(stat -c %s target/release/$b) bytes)"
```
flyq: **correctness** · `CONFIRMED`

`2>/dev/null || echo '<no --version>'` discards stderr and every non-zero exit alike, so a binary that segfaults, is linked against a missing library, or was built from the wrong version prints the same `<no --version>` as a binary that simply has no flag — the step stays green and the broken artifact is attached to the public Release.

<details>
<summary>Failure scenario</summary>

Any non-zero exit collapses to the same string: `target/release/stateless-validator` dying on a missing shared library prints `<no --version>  (12345678 bytes)`, the step exits 0, and `release-assets` uploads it. The size field is the only remaining signal, and it is satisfied by any file that exists.

As mega-maxwell's nit at L64 notes, the comment at lines 58-59 has this backwards — `stateless-validator` does carry `#[clap(author, version, ...)]`; it is `debug-trace-server` that does not. That is what makes the gate available on exactly the binary the comment says cannot have one. Asserting it also catches a tag / `Cargo.toml` version mismatch, which nothing else in the release flow checks at this point:

```bash
[[ "$(target/release/stateless-validator --version)" == "stateless-validator ${TAG#v}" ]]
```

Keeping `debug-trace-server` informational is fine; the point is that one of the two can be a real gate and currently neither is.

</details>

.github/workflows/on-release.yml
```diff
@@ -0,0 +1,71 @@
...
+
+permissions:
+  contents: read
+
+env:
+  TAG: ${{ github.ref_name }}
+  # A release event is never a dry run; a dispatch defaults to one.
+  DRY_RUN: ${{ github.event_name == 'workflow_dispatch' && inputs.dry_run || false }}
+
+jobs:
```
flyq, line 47: **correctness** · `CONFIRMED`

This is the only workflow in the repo without a `concurrency` group, and two runs for the same tag are reachable, both ending in a delete-then-upload of the same asset names.

<details>
<summary>Failure scenario</summary>

`build-and-test.yml:10`, `coverage.yml:10`, `release.yaml:13` and `release-tracing.yaml:13` all declare one; this job does not. Two runs for one tag are reachable — the `release: published` run plus a `workflow_dispatch` with `dry_run=false` on the same tag, or a re-run of a partially failed job. Both end in `gh release upload "$TAG" ... --clobber` inside `release-assets`, and `--clobber` is delete-then-upload: overlapping runs can race into a 422 "asset already exists", or leave an asset deleted by one run while the other is mid-upload, so the Release is briefly or permanently missing a binary.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: false
```

`github.ref` is the tag ref for both triggers, so it groups per tag. Note this must be `github.ref` and not `env.TAG`: the `env` context is not available to the workflow-level `concurrency` key (only `github`, `inputs`, `vars`). `cancel-in-progress` should stay `false` — the deliberate deviation from the siblings — since cancelling an upload mid-flight is the failure mode being avoided.

</details>

flyq (commented):
## Code review — `on-release.yml` (PR #207)

Reviewed at `88d5ca6d882c` · high effort · 3 findings

Reviewed `.github/workflows/on-release.yml` (the only file in this PR, 71 lines) at `88d5ca6`. One finding is blocking for v2.0.18 specifically — the workflow is correct, but it is not where the release that needs it will look for it.

### Inline findings

| # | Where | Finding |
|---|---|---|
| 1 | `.github/workflows/on-release.yml:15` | Won't run for v2.0.18: the file is not on the release branch |
| 2 | `.github/workflows/on-release.yml:62` | Version step swallows every failure; stays green on a broken binary |
| 3 | `.github/workflows/on-release.yml:32` | No concurrency group; two runs per tag race on --clobber upload |

### Checked and dropped

Things that looked like findings and are not, recorded so the next reader doesn't re-derive them:

- **No `token:` input on `release-assets`, and nothing generates `SHA256SUMS`.** The shared `megaeth-labs/.github/.github/actions/release-assets@main` generates the sums file itself and defaults `token` to `${{ github.token }}`. Both non-issues.
- **No `git_config` / `PAT_GAO_CI` step, which `release.yaml` needs.** All four git dependencies (`mega-evm`, `salt`, and the two reth repos) are public, and there are no submodules, so the credential step is genuinely unnecessary here.
- **`cargo build --bin stateless-validator --bin debug-trace-server` at a virtual-workspace root.** Resolves correctly — verified by probing with a bogus bin name.
- **The `&&` / `||` ternary in `DRY_RUN`.** Correct in all three cases; the fallback `false` is also the value the collapsed branch needs.
- **`timeout-minutes: 60`.** Generous, not tight: the `Build` step of the last five `release.yaml` runs — same tree, same `lto = "thin"` / `codegen-units = 1` / `cache: false` — took ~5.5 minutes each. The one run with a 64-minute wall clock spent that in the queue, not building.

.github/workflows/on-release.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: cargo build --release --locked --bin stateless-validator --bin debug-trace-server
+
+      - name: Report binary versions
+        # Informational: stateless-validator's CLI does not carry a clap
+        # version, so this cannot be a gate the way it is for mega-evme.
+        run: |
+          for b in stateless-validator debug-trace-server; do
+            echo "$b: $(target/release/$b --version 2>/dev/null || echo '<no --version>')  ($(stat -c %s target/release/$b) bytes)"
+          done
+
```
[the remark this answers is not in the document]
Troublor: Fixed in 89aedf4. The comment was wrong: `stateless-validator` does carry `#[clap(author, version, …)]`, so the step is now a gate on that binary (it must report `stateless-validator <tag version>`; warning on a dry run, hard stop on a release, the same as mega-evme). `debug-trace-server` is the one without a version flag; it is exercised with `--help` instead, and the step comment now says so.

.github/workflows/on-release.yml
Troublor: Agreed, and this is why #209 exists: it cherry-picks this workflow onto `release-v2.0.18` through its own reviewed PR (base `release-v2.0.18`, as the `release-*` ruleset requires) and merges before the settle PR #208. Settle is then re-run on the new tip, so the settle merge commit that gets tagged carries the file. 89aedf4 also adds the release-branch requirement and the rehearsal caveat to the header comment, as `release-publish.yml` does.

.github/workflows/on-release.yml
Troublor: Fixed in 89aedf4. `stateless-validator --version` is asserted against `stateless-validator ${TAG#v}`, mirroring the mega-evme gate (warning on a dry run, `exit 1` on a release). `debug-trace-server` has no version flag, so the step runs `debug-trace-server --help` under `set -euo pipefail`: a binary that cannot load or crashes fails the step instead of printing `<no --version>`. Sizes are still reported.

.github/workflows/on-release.yml
Troublor: Added in 89aedf4: `concurrency: { group: ${{ github.workflow }}-${{ github.ref }}, cancel-in-progress: false }`, with a comment explaining the per-tag grouping and why `cancel-in-progress` deliberately differs from the sibling workflows.

